PR #10, opened 2017-07-05
Added public compressedSize & uncompressedSize to Entry.

marcopifferi:
Hello,

I added two public variables in order to get the compressed and uncompressed size of an Entry.

Best
Marco

### DIFF
--- a/Sources/ZIPFoundation/Entry.swift
+++ b/Sources/ZIPFoundation/Entry.swift
@@ -132,6 +132,28 @@ public struct Entry: Equatable {
         }
         return checksum
     }
+    public var compressedSize: UInt32 {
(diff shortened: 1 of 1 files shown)


weichsel:
Hi Marco,
Sorry for the late reply.  

`Entry` only offers a bare minimum of metadata by design. By adding more public properties to the class we'd probably end up duplicating the whole state of the `CentralDirectoryStructure` and the `LocalFileHeader` in `Entry`. We could expose `uncompressedSize` via the `Entry.attributes` dictionary. This already contains the permissions and the modification time stamp. 
Foundation/`FileManager` has no notion for compression so it sadly doesn't offer a `FileAttributeKey` that we could use for `compressedSize`.

PR #14, opened 2017-07-12
Change file modes to unexecutable

thii:
Most of the files are in executable mode, which they shouldn't be.


weichsel:
I defined the default permissions in `Archive.swift:18` to be `0o755`. So maybe it'd make sense to change them to `0o644` there as well.

PR #16, opened 2017-07-12
Fixed some issues causing compile errors on Xcode 8.3.3

stephanheilner:


Sources/ZIPFoundation/Archive+Writing.swift
```diff
@@ -118,7 +118,7 @@ extension Archive {
         let centralDirectory = try self.writeCentralDirectoryStructure(localFileHeader: localFileHeader,
                                                                        relativeOffset: offset,
                                                                        externalFileAttributes: externalFileAttributes)
-        if startOfCentralDirectory > UINT32_MAX {
```
weichsel: This change would violate the ZIP specification. 
`Int.max` is platform dependent. It's 2^63 − 1 on arm64 and x86_64 and 2^31 - 1 on older ARM versions (for example iPhone 5 & 5c) and i386.  
ZIP (without the ZIP64 extension) restricts the start of the central directory to be within UInt32 range (2^32 - 1).  
So a better approach would be to use `Int(UInt32.max)` here.

Sources/ZIPFoundation/Archive+Writing.swift
```diff
@@ -228,15 +228,15 @@ extension Archive {
...
         var sizeWritten = 0
         var checksum = CRC32(0)
         while position < size {
-            let readSize = (Int(size) - position) >= bufferSize ? Int(bufferSize) : (Int(size) - position)
+            let readSize = (size - position) >= bufferSize ? bufferSize : (size - position)
             let entryChunk = try provider(Int(position), Int(readSize))
             checksum = entryChunk.crc32(checksum: checksum)
             sizeWritten += try Data.write(chunk: entryChunk, to: self.archiveFile)
-            position += Int(bufferSize)
+            position += bufferSize
```
weichsel: This introduces an overflow crash. 
If size is `UInt32.max` and `position` is `UInt32`, the last iteration crashes. This is caught by the test suite in `testCreateArchiveAddTooLargeUncompressedEntry`.
I assume you got "binary operator cannot be applied" warnings. Swift 4 seems to implicitly upcast `UInt32` to make it compatible with `Int` here. Maybe you can keep `position` as `Int` and cast the `UInt32` values where necessary to make it compile with older Swift versions.

PR #31, opened 2017-12-01
Fix for code sign error

### LINKED ISSUE #33: FRAMEWORK_SEARCH_PATHS interferes with Gatekeeper
We currently ship with a `FRAMEWORK_SEARCH_PATHS` setting that leads to Gatekeeper rejections during launch.

Steps to reproduce:
1. Integrate ZIP Foundation into a macOS sample app (e.g. using Carthage)
1. Embed ZIP Foundation into the sample app
1. Archive build the sample
1. Zip and upload it to a web server (This is important because the issue is not reproducible when the app is only locally launched)
1. Download the sample app and try to launch

priya87:
Fixes #33

Removed RunPath search paths and framework search paths which causes signature problem.

### DIFF
--- a/ZIPFoundation.xcodeproj/project.pbxproj
+++ b/ZIPFoundation.xcodeproj/project.pbxproj
@@ -403,13 +403,8 @@
 			isa = XCBuildConfiguration;
 			buildSettings = {
 				ENABLE_TESTABILITY = YES;
-				FRAMEWORK_SEARCH_PATHS = (
-					"$(inherited)",
-					"$(PLATFORM_DIR)/Developer/Library/Frameworks",
-				);
 				HEADER_SEARCH_PATHS = "$(inherited)";
 				INFOPLIST_FILE = ZIPFoundation.xcodeproj/ZIPFoundation_Info.plist;
-				LD_RUNPATH_SEARCH_PATHS = "$(TOOLCHAIN_DIR)/usr/lib/swift/macosx";
 				OTHER_LDFLAGS = "$(inherited)";
 				OTHER_SWIFT_FLAGS = "$(inherited)";
 				PRODUCT_BUNDLE_IDENTIFIER = ZIPFoundation;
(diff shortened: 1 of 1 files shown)


weichsel:
Thanks for your PR.

Can you provide more details about your setup. When did you encounter the signing issues?
How did you integrate ZIP Foundation into your workspace (Carthage, SPM, CocoaPods...)?

ZIP Foundation currently uses the default Xcode project template (with the exception of multiple deployment targets for iOS, macOS & watchOS) so signing should work automatically if the consuming project is set up properly.

priya87:
I used carthage and built with release configuration for macOS. I don't face any problem while signing and even no problem is shown during the verification of signature. But once I download the application via internet and try to open it, the gatekeeper shows a message that the app is from an unidentified developer. This is because  

> From 10.10.4, If  an app uses @rpath or an absolute path to link to a dynamic library outside of the app, the app will be rejected by Gatekeeper.

Thank you,

weichsel:
Thank you for the details - Great find. 
I can reproduce the problem when archiving a Test app that integrated the framework via Carthage.

I am not sure yet how those `FRAMEWORK_SEARCH_PATHS` and `LD_RUNPATH_SEARCH_PATHS` settings came to be. I either created the initial project in Xcode or generated it via `swift package generate-xcodeproj` and never touched those build settings. I will test your PR with the other 2 package managers and merge it over the weekend. 

Xcode 9.1 does indeed not set `FRAMEWORK_SEARCH_PATHS`, but it does set `LD_RUNPATH_SEARCH_PATHS` to `"$(inherited) @executable_path/../Frameworks @loader_path/Frameworks";`

So maybe we should try the default value and check if this also interferes with Gatekeeper.

priya87:
The problem is when the project is generated via `swift package generate-xcodeproj`

weichsel:
Yes. There's even a comment about that in the SPM source:
https://github.com/apple/swift-package-manager/blob/master/Sources/Xcodeproj/pbxproj().swift#L418

I merged your PR and re-added platform-dependent settings for `LD_RUNPATH_SEARCH_PATHS` (I used the default values that Xcode adds for macOS/iOS/tvOS/watchOS framework targets). 
With these changes, apps that integrate ZIP Foundation via Carthage now work for distribution. 
Other package managers should be not affected.

Thanks again for your contribution.

PR #53, opened 2018-03-07
 Fix an issue with unpacking files within directories on macOS

### LINKED ISSUE #52: Issue with Read-Only Directory inside Zip File
# Summary
A zip file I'm trying to unpack has a root level directory, which has write and executable bits set to false. This causes the unpack process to fail, because the process cannot write into the directory it has just created.

The solution I'd like to suggest is to manually set the `rwx` bits on all directories created during extraction.

# Steps to Reproduce
1. Download  https://github.com/KSP-CKAN/CKAN-meta/archive/master.zip
2. Unpack using the following code:

```swift
let workingDirectory = URL.init(fileURLWithPath: "/Users/yourname/zipfoundation_temp")
let sourceUrl = workingDirectory.appendingPathComponent(zipFileName)
let destinationUrl = workingDirectory

do {
    try fileManager.unzipItem(at: sourceUrl, to: destinationUrl)
} catch {
    print("Extraction of ZIP archive failed with error:\(error)")
}
```

# Expected Results
The ZIP file is extracted.

# Actual Results
The extraction fails with the following error: 

`Extraction of ZIP archive failed with error:Error Domain=NSCocoaErrorDomain Code=513 "You don’t have permission to save the file “10kmOmniEVA4RT” in the folder “CKAN-meta-master”." UserInfo={NSFilePath=/Applications/Kerbal Space Program/temp/CKAN-meta-master/10kmOmniEVA4RT, NSUnderlyingError=0x600000851e50 {Error Domain=NSPOSIXErrorDomain Code=13 "Permission denied"}}`

# Submitter's Note
Thanks for the excellent docs to this library :)

winsmith:
Fixes #52 

# Changes proposed in this PR

In this PR, we change the order of extraction to extract all directories first. We also apply different default permissions to directories (`755`) than we do to files (`644`). (Directories are extracted first so that the default directory attributes are set instead of the default file attributes.)

This is done by renaming the existing `defaultPermissions` setting to `defaultFilePermissions`, and introducing a new setting `defaultDirectoryPermissions` and updating the `FileManager.attributes:from:` method to take the entry type into account when returning default permissions. 

# Tests performed

- I verified that archives extracted with this tweaked algorithm have exactly the same file and directory permissions as archives extracted with macOS's `unzip` command line tool.
- I've only tested this thoroughly on macOS. A quick run on iOS seems to work though.
- The unit tests pass

Zip files with no defined permissions for files and directories (like the ones that are generated by GitHub's "download archive" feature) cause ZIPFoundation to use its default permissions.

The default permissions are set to `644` in `Archive.swift`. While this is enough for files to be readable and writable, `FileManager` throws a `Permission Denied` error when trying to write into a directory that is not explicitly marked as executable. This caused the error described in #52, which this PR aims to fix.

## Example Project

In the [zip-foundation-fix branch of my tiny Launch Pad project](https://github.com/winsmith/launch-pad/tree/feature/zip-foundation-fix) you'll find a copy of the files embedded in an Xcode project. Run that project, and it will download a zip file and try to extract it. (Output is on the console, please ignore the UI). 

# Open Issues

While Unit Tests pass, swiftlint complains about `Fallthrough Violation: Fallthrough should be avoided. (fallthrough)` on lines I did not touch.

### DIFF
--- a/Sources/ZIPFoundation/FileManager+ZIP.swift
+++ b/Sources/ZIPFoundation/FileManager+ZIP.swift
@@ -124,8 +124,9 @@ extension FileManager {
         }
     }
 
-    class func attributes(from centralDirectoryStructure: CentralDirectoryStructure) -> [FileAttributeKey: Any] {
-        var attributes = [.posixPermissions: defaultPermissions,
+    class func attributes(from entry: Entry) -> [FileAttributeKey: Any] {
+        let centralDirectoryStructure = entry.centralDirectoryStructure
+        var attributes = [.posixPermissions: entry.type == .directory ? defaultDirectoryPermissions : defaultFilePermissions,
                           .modificationDate: Date()] as [FileAttributeKey: Any]
         let versionMadeBy = centralDirectoryStructure.versionMadeBy
         let fileTime = centralDirectoryStructure.lastModFileTime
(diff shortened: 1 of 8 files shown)


weichsel:
Thank you for looking into this. 
I hope to find some time for a full review tomorrow, but your approach with different default permissions for directories looks good.  👌

The swiftlint messages regarding `fallthrough` are false positives IMO. I reported this some time ago and the swiftlint team already implemented an - yet unmerged - PR to fix this. (https://github.com/realm/SwiftLint/issues/1892)

Sources/ZIPFoundation/FileManager+ZIP.swift
```diff
@@ -124,8 +124,10 @@ extension FileManager {
         }
     }
 
-    class func attributes(from centralDirectoryStructure: CentralDirectoryStructure) -> [FileAttributeKey: Any] {
-        var attributes = [.posixPermissions: defaultPermissions,
+    class func attributes(from centralDirectoryStructure: CentralDirectoryStructure,
```
weichsel: This would be a good opportunity to refactor this method. We could rename it to `attributes(for entry: Entry)` and retrieve CDS, `entryType` and `osType` from the passed in `entry` struct.

PR #56, opened 2018-03-29
Use new function name to allocate bytes in Swift 4.1

danielctull:
Swift 4.1 warning message: 'allocate(bytes:alignedTo:)' is deprecated: renamed to 'allocate(byteCount:alignment:)'

I'm not sure if you want to maintain backwards compatibility for some time, so I changed this function name inside a check on the Swift version.


weichsel:
Thanks for your Pull Request, Daniel.

I will try to maintain Swift 4.0 compatibility at least until Swift 5.0 gets released - But this might change if Swift 5.0 introduces too many source breaking changes.

PR #68, opened 2018-06-02
Compression for FileManager zipItem

### LINKED ISSUE #66: The FileManager zipItem API is not compressing individual files
# Summary
Noticed that ZIP files were actually bigger than the text file I was zipping. 

# Steps to Reproduce
Call FileManager.zipItem with a large text file

# Expected Results
A Zip file that is smaller

# Actual Results
Larger zip file

# Regression & Version
Just started using ZipFoundation so do not know

# Related Link

I changed my copy to work. All that needs to change for individual files is to add a call parameter to Archive.addEntry

            progress?.totalUnitCount = archive.totalUnitCountForAddingItem(at: sourceURL)
            let baseURL = sourceURL.deletingLastPathComponent()
            try archive.addEntry(with: sourceURL.lastPathComponent, relativeTo: baseURL, **compressionMethod:.deflate**, progress: progress)

Based on this problem, it does not matter which Xcode or hardware platform is used. The default for compressionMethod is none. Maybe it would be better to change the default, but otherwise, you would need to change everywhere addEntry is called.

Adding this small change was enough to have a 500K text file go down to 50K zip. It was a 510K Zip file before.

weichsel:
Fixes #66 

# Changes proposed in this PR
* Adds a `compressionMethod` parameter to `zipItem`

### DIFF
--- a/Sources/ZIPFoundation/FileManager+ZIP.swift
+++ b/Sources/ZIPFoundation/FileManager+ZIP.swift
@@ -25,10 +25,12 @@ extension FileManager {
     ///   - destinationURL: The file URL that identifies the destination of the zip operation.
     ///   - shouldKeepParent: Indicates that the directory name of a source item should be used as root element
     ///                       within the archive. Default is `true`.
+    ///   - compressionMethod: Indicates the `CompressionMethod` that should be applied.
     ///   - progress: A progress object that can be used to track or cancel the zip operation.
     /// - Throws: Throws an error if the source item does not exist or the destination URL is not writable.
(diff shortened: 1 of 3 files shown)


phimage:
Hi, I like this PR 👍 
I forked the project and see that my needs are already fulfilled

A tagged released is planned? before swift 4.2?
If yes I will be able to update my project to use compression mode
https://github.com/phimage/FileZipKit

weichsel:
Hi Eric, 
Can't promise it 💯, but a tagged release (0.9.7) is planned before Swift 4.2 ships.

PR #76, opened 2018-08-19
Added Swift 3 support

colemancda:
Added Swift 3.0.2 ... 3.1.1 support (of course Swift 3.2 with Swift 4 compiler as well).

### DIFF
--- a/Package.pins
+++ b/Package.pins
@@ -1,0 +1,12 @@
+{
+  "autoPin": true,
+  "pins": [
+    {
+      "package": "CZlib",
+      "reason": null,
+      "repositoryURL": "https://github.com/IBM-Swift/CZlib.git",
+      "version": "0.1.2"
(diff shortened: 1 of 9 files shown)


weichsel:
Sorry for the delayed reply and thanks for your efforts.  

Sadly Swift is still evolving a lot (especially the tooling). This makes it very hard to guarantee quality across all supported platforms and versions. It's already quite difficult to cover all Apple platforms (macOS, iOS, tvOS & watchOS), Linux, 32bit/64bit _and_ the 2 versions of Foundation (Darwin & Swift Foundation). Adding backwards compatibility for an older major version of Swift makes the code more cluttered and doubles the effort necessary to build and test on all platform combinations.  

I really appreciate the work you put into this and will forward people that request Swift 3 compatibility to your fork.

PR #78, opened 2018-09-07
Support in-memory processing of archives 

microtherion:
Hi! I'm implementing a HTTP server that sometimes needs to extract data from uploaded dictionaries. Since these are ZIP encoded, I was looking for a library, and came across your project. However, I wanted to avoid writing out the uploaded data to a temp file, so I added a new method to your `Archive` class.

# Changes proposed in this PR
* Added initializer `Archive.init?(data: UnsafeMutableRawPointer, length: Int)` that creates a read-only archive from an in-memory pointer instead of an URL.

This is almost certainly not yet in a state where you'd consider merging it; I wanted to send you this to start a discussion on whether you'd be potentially interested in the change. See open issues below.

# Open Issues
* Code relies on `fmemopen` which was only introduced in macOS 10.13 / iOS 11.0
* Code only implements `.read` mode, because that's my only use case, and buffer handling for the other modes could get tricky.
* No unit tests yet.


weichsel:
This sounds like a great feature - Thanks for your contribution! 

> Code relies on fmemopen which was only introduced in macOS 10.13 / iOS 11.0

Making this macOS 10.13/iOS 11.0 only shouldn't be a big issue. We could just add an availability annotation to the added initializer. (I am not sure about the availability of `fmemopen` on Linux, but I assume it's in all newer glibc versions)

> Code only implements .read mode, because that's my only use case, and buffer handling for the other modes could get tricky.

Shipping only `.read` support works for me. Maybe we can find a way to communicate that via the API. (e.g. by using a convenience initializer named `makeArchiveReadingMemory(from ...)` or  by adding the `mode` parameter but asserting if someone passes `.write`)

> No unit tests yet.

Tests would be great 🙈

microtherion:
Thanks for your feedback! I will try to:

* Add availability annotations on Apple platforms (It appears the calls have been around on Linux for a [long time](http://man7.org/linux/man-pages/man3/fmemopen.3.html)).
* Add unit tests.
* Think again about supporting write mode (I've run into an use case, and an implementation strategy using `open_memstream`).

microtherion:
Thanks for your patience! I now have a new iteration of the design:

```
// Raw methods using UnsafePtr

// Create an in-memory, read-only archive from the provided Data
Archive.init?(readData: UnsafeRawPointer, length: Int)
// Create an in-memory, write-only archive
init?(writeData: UnsafeMutablePointer<UnsafeMutablePointer<Int8>?>, length: UnsafeMutablePointer<Int>)

// Convenience wrappers using closures

// Turn contents of Data into read-only archive and extract within closure
static Archive.extract(data: Data, body: (Archive) -> Void) throws
// Create write-only archive, populate within closure, and return resulting Data
static Archive.dataWithArchive(body: (Archive) -> Void) throws -> Data
```

I've added one unit test for each of these functions. All unit tests pass on macOS (What's the easiest way to test on Linux?)

# Open issues
* The two sets of methods each have their advantages and disadvantages. The `RawPointer` based ones are more flexible, but uglier. The wrapper methods are far more elegant, but all manipulation has to be confined to their closures. Would you like to keep both sets public, or would it be preferable to make the former private?
* The write-only archive is truly write-only. Any reading back is impossible. This causes read-based methods to fail in ways which may be surprising to users. Is it worth introducing a .writeOnly archive state so these methods will fail early?

Let me know what you think.

weichsel:
Thank you for your follow up and sorry for the delayed reply.  
The additional write support looks really useful - looking forward to use that myself.  

Regarding the API, I always aimed for simplicity over flexibility when designing the public interface. So I'd prefer just one initializer for in-memory archives.  
Ideally (disclaimer: I have very limited experience with `fmemopen` and `open_memstream`), our API should abstract away the `withUnsafeBytes` dance that Swift requires when dealing with pointers. I don't know if that's achievable, but I think as API user, the most convenient way to deal with in-memory archives would be to just have one additional initializer similar to the `URL`-based one, that takes a `Data` argument.  
e.g.: `Archive(data: Data, mode: AccessMode)`

Thank you for the additional tests! To exercise the test suite on Linux, I always use a Docker image:
`docker pull norionomura/swift:latest`

Run test suite:
`docker run --rm --volume "$(pwd):/package" --workdir "/package" norionomura/swift:latest /bin/bash -c "apt-get -y update && apt-get install -y libz-dev && swift package resolve && swift test -c release -Xswiftc -enable-testing --build-path ./.build/linux"`

The suite doesn't run through due to missing `Swift Foundation` API, but it works for most test cases.

weichsel:
Hello Matthias,

I finally found some time to look into your implementation. (Actually it took me several attempts, because the write part is really tricky - I think I now understand the things you pointed out.). 

While experimenting with your unit tests (thanks for providing them!), I think I found a show stopper with `open_memstream`: According to [this](https://stackoverflow.com/questions/30642450/can-i-read-stream-produced-by-open-memstream), the returned stream is strictly write-only.  
As ZIP Foundation relies on several read operations during `addEntry:`,  this leads to subtle errors when performing > 1 `add` operations.  `addEntry:` stores away the existing central directory before adding new entries. This works for the case where the currently added entry is the first one in the archive, but it fails if it has to perform a read to copy existing CD structures.  
You can reproduce this by adding the following to your `testCreateArchiveAddUncompressedEntryToMemory` test case:
```swift
let relativePath2 = "xy.png"
let tmpData = "test".data(using: .utf8)!
try archive.addEntry(with: relativePath2, type: .file, uncompressedSize: UInt32(tmpData.count), provider: { (_, _) -> Data in
    return tmpData
})
```

Currently I can't think of a simple solution that would work around that limitation.  

For read mode, it shouldn't be a problem to remove `final` from `Archive` and introduce an `InMemoryArchive` subclass that provides storage for `data`. 
As far as I could see it would work to pass the archive init a data object and then open the file handle via: 
```swift
self.archiveFile = data.withUnsafeBytes({ (ptr) -> UnsafeMutablePointer<FILE> in
    return fmemopen(UnsafeMutableRawPointer(mutating: ptr), data.count, "rb")
})
```

microtherion:
Thanks for the feedback! I'll think about how to address the write bug (and I better include a unit test with more than one entry next time!).
My understanding of Swift memory safety is rather limited, but my understanding is that `ptr` in an `withUnsafeBytes` block is not suppose to outlive the scope of that block, which in your example it indirectly does.

weichsel:
OK, no need to split this PR then. I just suggested that because I thought you might need the in-memory read support in your project and a tagged release would be easier to manage as dependency.  

I made another attempt to solve the write problem recently, but without luck so far.   
I think that working around the write-only limitation of `open_memstream` will require some refactoring in the `add`/`removeEntry` methods. Probably using a dedicated archive `FILE`  handle for read operations in `add/remove` would be enough.

microtherion:
Withdrawing this in favor of the much more elegant #143.

PR #79, opened 2018-09-13
Swift 4.2 support for Swift Package Manager

trilorez:
Adds Swift 4.2 support for Swift Package Manager.
Note: when attempting to run tests, I get stuck at `testCreateArchiveAddLargeCompressedEntry`, but this occurs before any of my changes were made.

### DIFF
--- a/Package@swift-4.2.swift
+++ b/Package@swift-4.2.swift
@@ -1,0 +1,21 @@
+// swift-tools-version:4.2
+import PackageDescription
+
+#if os(macOS) || os(iOS) || os(watchOS) || os(tvOS)
+let dependencies: [Package.Dependency] = []
+#else
+let dependencies: [Package.Dependency] = [.package(url: "https://github.com/IBM-Swift/CZlib.git", .exact("0.1.2"))]
+#endif
+
+let package = Package(
+    name: "ZIPFoundation",
+    products: [
+        .library(name: "ZIPFoundation", targets: ["ZIPFoundation"])
+    ],
+	dependencies: dependencies,
+    targets: [
+        .target(name: "ZIPFoundation"),
+		.testTarget(name: "ZIPFoundationTests", dependencies: ["ZIPFoundation"])
+    ],
+    swiftLanguageVersions: [.v4, .v4_2]
+)


weichsel:
Thank you for your contribution.  

Wouldn't lead bumping the `swift-tools-version` to problems for people that still use `4.0`,`4.1` toolchains?

The tests aren't stuck, they are just very slow when the framework is built without optimizations (= `Debug`). To run the test suite with `release` configuration you can use the following:
- macOS: `swift test -c release -Xswiftc "-target" -Xswiftc "x86_64-apple-macosx10.12" -Xswiftc -enable-testing` 
- Linux: `swift test -c release -Xswiftc -enable-testing` 
If you are working on Linux, you can leave out the

trilorez:
Good point, I will revise my pr to support `swift-tools-version 4.0` (it looks like there isn't a `4.1`) using [version specific manifest selection](https://github.com/apple/swift-package-manager/blob/master/Documentation/Usage.md#version-specific-manifest-selection).

Got it regarding the tests.

weichsel:
Great, thank you!

trilorez:
The pr has been updated to be purely additive

weichsel:
Thanks for the update. But as the package is not using anything that's version specific, we probably should avoid version specific manifest selection: 
From the docs: 

> It is not expected the packages would ever use this feature unless absolutely necessary to support existing clients.

Wouldn't it be sufficient to leave the `swift-tools-version` as-is and just add the `v4_2` tag to `swiftLanguageVersions`?

trilorez:
Unfortunately `v4_2` was added in [swift-tools-version 4.2](https://github.com/apple/swift-package-manager/blob/master/Documentation/PackageDescriptionV4_2.md#swift-language-version). Before `swiftLanguageVersions` was just an [int array](https://github.com/apple/swift-package-manager/blob/master/Documentation/PackageDescriptionV4.md#swiftlanguageversions). Maybe a better path forward is to have the new manifest be named `Package.swift` and the old one as `Package@swift-4.swift` since 4.2 seems more forward compatible?

weichsel:
Adding an additional file adds another thing that must be taken into account when maintaining the package versions (e.g. when ending support for a specific Swift version).
As the package manifest file supports `#if` directives, we maybe can get away something like `#if swift(>=4.2)` and return the new Swift language version type and an `Int` for earlier versions.  
I am still not sure if we need to indicate Swift 4.2 support at all. The library does not use any 4.2 specific features. So it should work fine as-is for people using 4.2 toolchains.

PR #81, opened 2018-11-12
Allow use in app extensions

### LINKED ISSUE #75: Incompatible with extensions such as an iMessage app extension
# Summary
Including in a project to build an iMessage extension gets a linker warning _linking against a dylib which is not safe for use in application extensions_

This would result in app store rejection.

# Steps to Reproduce
- Add the framework using Carthage to an iMessage extension project in XCode 9.4.1
- build and check linker errors

# Expected Results
no errors

# Actual Results
linker warning `linking against a dylib which is not safe for use in application extensions: /XXXXXX/Carthage/Build/iOS/ZIPFoundation.framework/ZIPFoundation`

# Regression & Version
Current version 0.9.6 as per `cartfile.resolved`

# Related Link
- Flint project issue [111](https://github.com/MontanaFlossCo/Flint/issues/111)

rileytestut:
Fixes #75

Pretty small change. This allows ZIPFoundation to be imported and used from iOS app extensions. ZIPFoundation is already only using app extension-safe APIs, so it just required enabling this in the project settings.

# Changes proposed in this PR
* Allows use in app extensions by checking "Allow app extension API only" in framework's project settings

# Tests performed
* All tests still pass

### DIFF
--- a/ZIPFoundation.xcodeproj/project.pbxproj
+++ b/ZIPFoundation.xcodeproj/project.pbxproj
@@ -405,6 +405,7 @@
 		OBJ_45 /* Debug */ = {
 			isa = XCBuildConfiguration;
 			buildSettings = {
+				APPLICATION_EXTENSION_API_ONLY = YES;
 				CURRENT_PROJECT_VERSION = 0.9.6;
 				ENABLE_TESTABILITY = YES;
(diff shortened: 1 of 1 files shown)


weichsel:
Thanks for your contribution - Looks good.

There was already an issue for this (I linked it at the top of your PR).

PR #85, opened 2018-12-06
Added iOS target

colemancda:
To properly support Carthage for iOS

Fixes Carthage only building macOS framework.

### DIFF
--- a/ZIPFoundation.xcodeproj/project.pbxproj
+++ b/ZIPFoundation.xcodeproj/project.pbxproj
@@ -7,6 +7,13 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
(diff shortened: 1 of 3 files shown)


weichsel:
Hi,
Thank you for the pull request.  

ZIPFoundation uses an .xcodeproj with a single target for all platforms. This should work as-is with Carthage, but there were some Carthage versions that only built the macOS target of ZIPFoundation and then canceled the build because it was unable to build the watchOS target. This should be fixed in Carthage versions > 0.31.0.  
We also recommend using Carthage with the `--no-build` flag for ZIPFoundation.  
While this requires you to manually drag the ZIPFoundation.xcodeproj to your workspace, it leaves the build process to Xcode. In my experience, this leads to fewer platform dependent build problems.

PR #86, opened 2018-12-26
Remove FileAttributeKeys not supported in swift-corelibs when unzipping file

### LINKED ISSUE #65: Fatal Error: NSFileModificationDate not implemented when unzip item on Linux.
# Summary

Fatal error when tried to unzip on linux.
Same code, everything works well on macOS, but failed on linux.

# Steps to Reproduce

```
try FileManager.default.unzipItem(at: urlToZip, to: urlToDest)
```

I make a small demo to reproduce this:

```
git clone https://gitlab.com/isaacxen/zipFoundationErrorDemo.git
cd zipFoundationErrorDemo/
cp example.epub /tmp
swift package update
swift run
```

# Expected Results

successfully unzip.

# Actual Results

```
Fatal error: Attribute type not implemented: FileAttributeKey(rawValue: "NSFileModificationDate"): file Foundation/FileManager.swift, line 139
```

# Regression & Version

ZipFoundation: 0.9.6
Ubuntu 16.04
Swift 4.1

# Related Link

andrewtheis:
Fixes #65

# Changes proposed in this PR
* Temporary workaround for attribute not supported on Linux.

### DIFF
--- a/Sources/ZIPFoundation/Archive+Reading.swift
+++ b/Sources/ZIPFoundation/Archive+Reading.swift
@@ -53,8 +53,13 @@ extension Archive {
             }
             checksum = try self.extract(entry, bufferSize: bufferSize, progress: progress, consumer: consumer)
         }
-        let attributes = FileManager.attributes(from: entry)
+
+        var attributes = FileManager.attributes(from: entry)
+        #if os(Linux) // Certain keys are not yet supported in swift-corelibs
+        attributes.removeValue(forKey: .modificationDate)
+        #endif
         try fileManager.setAttributes(attributes, ofItemAtPath: url.path)
+        
         return checksum
     }
 


weichsel:
Thank you for your contribution and sorry for the delayed reply.  

I think this is a good workaround until corelibs Foundation gets `fileModificationDate` support.  
Probably we should change the condition to check for Apple platforms (where Darwin Foundation is used). This would make sure that other Swift platforms that have to rely on corelibs Foundation also get the fix:

```swift
#if os(macOS) || os(iOS) || os(watchOS) || os(tvOS)
#else // Certain keys are not yet supported in swift-corelibs
    attributes.removeValue(forKey: .modificationDate)
#endif
```

pvieito:
Even cleaner: 

```
#if !canImport(Darwin) // Certain keys are not yet supported in swift-corelibs
    attributes.removeValue(forKey: .modificationDate)
#endif
```

weichsel:
@pvieito We currently still support Swift 4 and `canImport` is a Swift 4.1 feature

PR #87, opened 2019-01-16
Fix traversal attack vulnerability

### LINKED ISSUE #74: Vulnerability of traversal attack
# Summary
ZIPFoundation could be attacted by archive traversal attack[1]

# Steps to Reproduce
1. Put `evil.zip`[2] attached to somewhere (directory A)
2. Unzip `evil.zip` using ZIPFoundation to another place (directory B).

# Expected Results
`evilfile` is extracted in directory B (or failed maybe)

# Actual Results
`evilfile` will be extracted in the parent directory of `evil.zip` (parent of directory A)

# Regression & Version
`0.9.6`

# Related Link
[1] https://en.wikipedia.org/wiki/Directory_traversal_attack#Zip/archive_traversal_attacks
[2] [evil.zip](https://github.com/weichsel/ZIPFoundation/files/2248247/evil.zip)

waynezhang:
Fixes #74 

# Changes proposed in this PR
* Fail extract if any entry tries to extract itself outside of destination path.

# Tests performed
All test cases + `testTraversalAttack`

# Open Issues
https://github.com/weichsel/ZIPFoundation/issues/74


weichsel:
Thank you for your contribution - especially for taking the time to also provide a test case!

I just found one issue that I don't fully understand yet: Is the removal of the last path component a measure for the edge case where the bad entry name is the same as the last path component of the destination directory?

Maybe a better way to handle this would be to ensure that the destination path has a `/` suffix before comparing. (Sadly this is a bit tricky because ZIP Foundation doesn't enforce that the passed in destination URL is a directory URL and additionally the standard Swift `URL` implicitly removes trailing path separators).  

One way to handle that would be the following extension to URL:

```(swift)
extension URL {
    func isContained(in parentDirectoryURL: URL) -> Bool {
        // Ensure that the passed in URL is really a directory URL
        let directoryName = parentDirectoryURL.lastPathComponent
        var directoryURL = parentDirectoryURL.deletingLastPathComponent()
        directoryURL = directoryURL.appendingPathComponent(directoryName, isDirectory: true)
        directoryURL = directoryURL.standardizedFileURL
        // Use Core Foundation URLs here because they keep the trailing path separator
        let destinationDirectoryPath = CFURLCopyPath(directoryURL as CFURL) as String
        let path = CFURLCopyPath(self as CFURL) as String
        return path.hasPrefix(destinationDirectoryPath)
    }
}
``` 

I am not sure if I thought this through correctly - so please let me know what you think about that extension.

waynezhang:
Yes the `deletingLastPathComponent` part is only for the edge case. But you are right that I didn't think it though. The case like `../../destinationURLxxx/xxx` will be ignored.
And based on your idea I think it's simpler to do it this way. How do you think about it?

```swift
extension URL {
    func isContained(in parentDirectoryURL: URL) -> Bool {
        let parentDirectoryURL = URL(
			fileURLWithPath: parentDirectoryURL.absoluteString,
			isDirectory: true
		).standardizedFileURL
        return self.absoluteString.hasPrefix(parentDirectoryURL.absoluteString)
    }
}
```

weichsel:
Great find! - I didn't think about `absoluteString`.
This is way more elegant than my CF version.  
There's a small bug: We should use `absoluteString` instead of `absolutePath` in the URL initializer. Otherwise we get multiple `file://` prefixes:

```swift
func isContained(in parentDirectoryURL: URL) -> Bool {
    // Make sure that the passed in URL is a directory URL (ensuring the `/` path suffix)
    let parentDirectoryURL = URL(fileURLWithPath: parentDirectoryURL.path, isDirectory: true).standardizedFileURL
    return self.absoluteString.hasPrefix(parentDirectoryURL.absoluteString)
}
```

waynezhang:
@weichsel Yeah I just found that and updated it :) I'm gonna fix this.

weichsel:
Thank you - I think we can merge this when you are finished.

waynezhang:
Just push a new fix on this.

PR #117, opened 2019-05-02
Swift 5 support for podspec and Xcode project file

fassko:
Fixes #

# Changes proposed in this PR
* Swift 5 support for podspec and Xcode project file

# Tests performed

Nothing else has been changed.

# Open Issues

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,16 @@
 # Changelog
 
+## [0.9.9](https://github.com/weichsel/ZIPFoundation/releases/tag/0.9.10)
+
+### Updated
+ - Swift 5.0 support in Xcode project and podspec file
+
 ## [0.9.9](https://github.com/weichsel/ZIPFoundation/releases/tag/0.9.9)
 
 ### Added
  - Swift 5.0 support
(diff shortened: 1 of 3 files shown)


weichsel:
Hi,

Thank you for your pull request.  

The Swift 4.0 language version in Xcode/CocoaPods is by design.  
Xcode only supports one `SWIFT_VERSION` setting per project. As the Swift 5.0 toolchain can compile in Swift 4.x mode, I decided to leave the language version at the lowest supported setting to keep the project compilable for people that integrate via Carthage and are still using Swift versions < 5.0.  

CocoaPods will start to support multiple Swift versions with the upcoming 1.7 release ([source](http://blog.cocoapods.org/CocoaPods-1.7.0-beta/)) and I plan to add support for that in the next ZIP Foundation release.  

I know that the current situation is not ideal as only users consuming the framework via Swift Package manager get the Swift 5.0 version, but I couldn't find a solution that would cause Xcode to fall back to older Swift versions if the current toolchain is not Swift 5 or newer.

fassko:
@weichsel Thanks for the explanation. I think a good way is to create a branch or tag for older Swift versions. But I agree that cocoapods 1.7 will be a big help in that regard.

PR #119, opened 2019-05-09
Levi/108

### LINKED ISSUE #108: Performance improvement: consider using FileManager.default.url(for:,in:,appropriateFor:, create:) instead of NSTemporaryDirectory()
**Suggestion:**

Performance of removing an entry could be improved a bit for archives that reside on volumes, other than the main volume (i.e. the volume of the user's home directory) by using a temporary directory on the same external volume and thus avoiding copying the temporary archive across volumes.

**Details:**

`NSTemporaryDirectory()` always returns the path of the temporary directory for the current user. If a zip archive is on a different volume, the performance of removing an entry could be improved by using `FileManager.url(for:in:appropriateFor:create:)` instead. The FileManager method _returns a temporary directory on the same volume as the target file_ so that replacing the original with the temporary new version of the archive can be achieved in one step.

This is also the recommended way by Apple: https://developer.apple.com/documentation/foundation/1409211-nstemporarydirectory

**Code:**

In Archive+Writing.swift, around line 150 in `Archive.remove()`, replace:

    let tempArchiveURL =  URL(fileURLWithPath: NSTemporaryDirectory()).appendingPathComponent(uniqueString)

with: 

    let tempArchiveURL = try FileManager().url(for: .itemReplacementDirectory, in: .userDomainMask, appropriateFor: self.url, create: true).appendingPathComponent(uniqueString)

https://github.com/weichsel/ZIPFoundation/blob/92d56accedbd385ae39d3521a46524ea07cab008/Sources/ZIPFoundation/Archive%2BWriting.swift#L150

levigroker:
Fixes #108 

# Changes proposed in this PR
* Using modern temporary directory mechanism, with fallback if needed.
* Addresses archive replacement issue when archives reside on different volumes.
* Removed `fileExists` check when creating directory as this is contrary to recommended practice.

# Tests performed
* Added test for temporary directory creation mechanism.
* Added test for cross-link (separate volume) archive replacement changes.

* Opened up `private` methods to `internal` (implicit) for testing purposes.
* Maintained 100% code coverage and no swiftlint issues as per contribution guidelines.

### DIFF
--- a/Sources/ZIPFoundation/Archive+Writing.swift
+++ b/Sources/ZIPFoundation/Archive+Writing.swift
@@ -327,7 +327,7 @@ extension Archive {
         _ = try Data.write(chunk: endOfCentralDirRecord.data, to: self.archiveFile)
     }
 
-    private func replaceCurrentArchiveWithArchive(at URL: URL) throws {
+    func replaceCurrentArchiveWithArchive(at URL: URL) throws {
         fclose(self.archiveFile)
         let fileManager = FileManager()
         #if os(macOS) || os(iOS) || os(watchOS) || os(tvOS)
--- a/Sources/ZIPFoundation/Archive+Writing.swift
+++ b/Sources/ZIPFoundation/Archive+Writing.swift
@@ -146,7 +146,7 @@ extension Archive {
     ///   - progress: A progress object that can be used to track or cancel the remove operation.
     /// - Throws: An error if the `Entry` is malformed or the receiver is not writable.
     public func remove(_ entry: Entry, bufferSize: UInt32 = defaultReadChunkSize, progress: Progress? = nil) throws {
-        let tempDir = uniqueTemporaryDirectoryURL()
+        let tempDir = self.uniqueTemporaryDirectoryURL()
         defer { try? FileManager().removeItem(at: tempDir) }
         do { try FileManager().createParentDirectoryStructure(for: tempDir) } catch { throw ArchiveError.unwritableArchive }
         let uniqueString = ProcessInfo.processInfo.globallyUniqueString
--- a/Sources/ZIPFoundation/Archive+Writing.swift
+++ b/Sources/ZIPFoundation/Archive+Writing.swift
@@ -146,12 +146,14 @@ extension Archive {
     ///   - progress: A progress object that can be used to track or cancel the remove operation.
     /// - Throws: An error if the `Entry` is malformed or the receiver is not writable.
     public func remove(_ entry: Entry, bufferSize: UInt32 = defaultReadChunkSize, progress: Progress? = nil) throws {
+		let manager = FileManager()
         let tempDir = self.uniqueTemporaryDirectoryURL()
-        defer { try? FileManager().removeItem(at: tempDir) }
-        do { try FileManager().createParentDirectoryStructure(for: tempDir) } catch {
-            throw ArchiveError.unwritableArchive }
-        let uniqueString = ProcessInfo.processInfo.globallyUniqueString
-        guard let tempArchive = Archive(url: tempDir.appendingPathComponent(uniqueString), accessMode: .create) else {
+        defer { try? manager.removeItem(at: tempDir) }
+		let uniqueString = ProcessInfo.processInfo.globallyUniqueString
+		let tempArchiveURL =  tempDir.appendingPathComponent(uniqueString)
+        do { try manager.createParentDirectoryStructure(for: tempArchiveURL) } catch {
+			throw ArchiveError.unwritableArchive }
+        guard let tempArchive = Archive(url: tempArchiveURL, accessMode: .create) else {
             throw ArchiveError.unwritableArchive
         }
         progress?.totalUnitCount = self.totalUnitCountForRemoving(entry)
--- a/Sources/ZIPFoundation/Archive+Writing.swift
+++ b/Sources/ZIPFoundation/Archive+Writing.swift
@@ -206,6 +206,24 @@ extension Archive {
             ProcessInfo.processInfo.globallyUniqueString)
     }
 
+    func replaceCurrentArchiveWithArchive(at URL: URL) throws {
+        fclose(self.archiveFile)
+        let fileManager = FileManager()
+        #if os(macOS) || os(iOS) || os(watchOS) || os(tvOS)
+        do {
+            _ = try fileManager.replaceItemAt(self.url, withItemAt: URL)
+        } catch {
+            _ = try fileManager.removeItem(at: self.url)
+            _ = try fileManager.moveItem(at: URL, to: self.url)
+        }
+        #else
+        _ = try fileManager.removeItem(at: self.url)
+        _ = try fileManager.moveItem(at: URL, to: self.url)
+        #endif
+        let fileSystemRepresentation = fileManager.fileSystemRepresentation(withPath: self.url.path)
+        self.archiveFile = fopen(fileSystemRepresentation, "rb+")
+    }
+
     private func writeLocalFileHeader(path: String, compressionMethod: CompressionMethod,
                                       size: (uncompressed: UInt32, compressed: UInt32),
                                       checksum: CRC32,
@@ -330,23 +348,5 @@ extension Archive {
         fseek(self.archiveFile, localFileHeaderStart, SEEK_SET)
         _ = try Data.write(chunk: existingCentralDirectoryData, to: self.archiveFile)
         _ = try Data.write(chunk: endOfCentralDirRecord.data, to: self.archiveFile)
-    }
-
-    func replaceCurrentArchiveWithArchive(at URL: URL) throws {
-        fclose(self.archiveFile)
-        let fileManager = FileManager()
-        #if os(macOS) || os(iOS) || os(watchOS) || os(tvOS)
-            do {
-                _ = try fileManager.replaceItemAt(self.url, withItemAt: URL)
-            } catch {
-                _ = try fileManager.removeItem(at: self.url)
-                _ = try fileManager.moveItem(at: URL, to: self.url)
-            }
-        #else
-            _ = try fileManager.removeItem(at: self.url)
-            _ = try fileManager.moveItem(at: URL, to: self.url)
-        #endif
-        let fileSystemRepresentation = fileManager.fileSystemRepresentation(withPath: self.url.path)
-        self.archiveFile = fopen(fileSystemRepresentation, "rb+")
     }
 }
--- a/Sources/ZIPFoundation/Archive+Writing.swift
+++ b/Sources/ZIPFoundation/Archive+Writing.swift
@@ -146,8 +146,12 @@ extension Archive {
     ///   - progress: A progress object that can be used to track or cancel the remove operation.
     /// - Throws: An error if the `Entry` is malformed or the receiver is not writable.
     public func remove(_ entry: Entry, bufferSize: UInt32 = defaultReadChunkSize, progress: Progress? = nil) throws {
-        let tempArchiveURL = createTempArchiveURL()
-        guard let tempArchive = Archive(url: tempArchiveURL, accessMode: .create) else {
+        let tempDir = uniqueTemporaryDirectoryURL()
+        defer {
+            try? FileManager().removeItem(at: tempDir)
+        }
+        let uniqueString = ProcessInfo.processInfo.globallyUniqueString
+        guard let tempArchive = Archive(url: tempDir.appendingPathComponent(uniqueString), accessMode: .create) else {
             throw ArchiveError.unwritableArchive
         }
         progress?.totalUnitCount = self.totalUnitCountForRemoving(entry)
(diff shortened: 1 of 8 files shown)


levigroker:
Seems some linting issues with the tests... is that something you're concerned with?
Additionally, I have no idea about what's wrong or how to address the linux failures.

weichsel:
Hi Levi,

Thanks for your contribution. I just had a brief look over your changes and they look 👌. 
I hope I have some time this weekend for a more in-depth review.  
Special thanks for taking the time to cover your additions with tests.  
The Linux build failures seem to be related to the `autoreleasepool` you are using in one of the test cases. APs are not available on Linux.

Sources/ZIPFoundation/Archive.swift
```diff
@@ -152,6 +152,7 @@ public final class Archive: Sequence {
             self.endOfCentralDirectoryRecord = endOfCentralDirectoryRecord
         case .create:
             guard !fileManager.fileExists(atPath: url.path) else { return nil }
+            do { try fileManager.createParentDirectoryStructure(for: url) } catch { return nil }
```
weichsel: If I understand this correctly, this was added to implicitly create intermediate directories. I'd prefer the previous behaviour, where archive creation fails if intermediate directories are not present.  
While this might be less convenient in some cases, it avoids unintentional folder creation and it informs the API user about potential problems with the provided path.  

The existing behaviour is also more in-line with Foundation APIs like `FileManager` where missing intermediate directories also result in creation errors: 
```swift
let fm = FileManager()
let tempURL = URL(fileURLWithPath: NSTemporaryDirectory())
let didCreate = fm.createFile(atPath: tempURL.appendingPathComponent(UUID().uuidString).path, contents: nil, attributes: nil)
let nestedTempURL = tempURL.appendingPathComponent("test")
let didCreateNested = fm.createFile(atPath: nestedTempURL.appendingPathComponent(UUID().uuidString).path, contents: nil, attributes: nil)
print("didCreate:\(didCreate) didCreateNested:\(didCreateNested)")
```

Sources/ZIPFoundation/FileManager+ZIP.swift
```diff
@@ -130,9 +130,7 @@ extension FileManager {
 
     func createParentDirectoryStructure(for url: URL) throws {
         let parentDirectoryURL = url.deletingLastPathComponent()
-        if !self.fileExists(atPath: parentDirectoryURL.path) {
-            try self.createDirectory(at: parentDirectoryURL, withIntermediateDirectories: true, attributes: nil)
-        }
+        try self.createDirectory(at: parentDirectoryURL, withIntermediateDirectories: true, attributes: nil)
```
weichsel: The `fileExists` check is there to silently return in case folder creation is not necessary. .

Tests/ZIPFoundationTests/ZIPFoundationWritingTests.swift
```diff
@@ -329,4 +329,100 @@ extension ZIPFoundationTests {
...
+    func testUniqueTemporaryDirectoryURL() {
+        let archive = self.archive(for: #function, mode: .create)
+        var tempURLs = Set<URL>()
+        defer {
+            for url in tempURLs {
+                try? FileManager.default.removeItem(at: url)
+            }
+        }
+        // We choose 2000 temp directories to test workaround for http://openradar.appspot.com/50553219
+        for _ in 1...2000 {
```
weichsel: Nice find 🙂. 
I saw that you already got a reply - quite fast for a Radar in pre-WWDC season.

Tests/ZIPFoundationTests/ZIPFoundationWritingTests.swift
```diff
@@ -329,4 +329,100 @@ extension ZIPFoundationTests {
...
+        defer {
+            try? FileManager.default.removeItem(at: tempDir)
+        }
+
+        let volName = "Test_\(UUID().uuidString)"
+        let dmgURL = tempDir.appendingPathComponent(volName).appendingPathExtension("dmg")
+        let script = """
+        #!/bin/bash
+        hdiutil create -size 5m -fs HFS+ -type SPARSEBUNDLE -ov -attach -volname "\(volName)" "\(dmgURL.path)"
+
```
weichsel: Great idea to create a multiple volume environment!

Sources/ZIPFoundation/Archive+Writing.swift
```diff
@@ -189,6 +192,15 @@ extension Archive {
 
     // MARK: - Helpers
 
+    func uniqueTemporaryDirectoryURL() -> URL {
+        if let tempDir = try? FileManager().url(for: .itemReplacementDirectory, in: .userDomainMask,
+                                                appropriateFor: self.url, create: true) {
+            return tempDir
+        }
+
+        return URL(fileURLWithPath: NSTemporaryDirectory()).appendingPathComponent(UUID().uuidString)
```
weichsel: I use `ProcessInfo.processInfo.globallyUniqueString` instead of `UUID` at several other source locations. While `globallyUniqueString` does not produce canonical UUIDs, it returns a string that is guaranteed to be unique for a network (because it uses the hostname as salt).

Tests/ZIPFoundationTests/ZIPFoundationWritingTests.swift
```diff
@@ -329,4 +329,100 @@ extension ZIPFoundationTests {
...
+        let archive = self.archive(for: #function, mode: .create)
+        var tempURLs = Set<URL>()
+        defer {
+            for url in tempURLs {
+                try? FileManager.default.removeItem(at: url)
+            }
+        }
+        // We choose 2000 temp directories to test workaround for http://openradar.appspot.com/50553219
+        for _ in 1...2000 {
+            autoreleasepool {
```
weichsel: This is the line that currently breaks the Linux build.  
Do we need the ARP here?

Sources/ZIPFoundation/Archive+Writing.swift
```diff
@@ -315,11 +327,16 @@ extension Archive {
...
+    func replaceCurrentArchiveWithArchive(at URL: URL) throws {
         fclose(self.archiveFile)
         let fileManager = FileManager()
         #if os(macOS) || os(iOS) || os(watchOS) || os(tvOS)
-            _ = try fileManager.replaceItemAt(self.url, withItemAt: URL)
+            do {
+                _ = try fileManager.replaceItemAt(self.url, withItemAt: URL)
+            } catch {
+                _ = try fileManager.removeItem(at: self.url)
+                _ = try fileManager.moveItem(at: URL, to: self.url)
```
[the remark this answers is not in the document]
weichsel: We should try to keep this method private because it changes internal state. I assume you exposed it to make it testable. Maybe we can exercise this codepath by triggering the calling method from the test case?

Sources/ZIPFoundation/Archive+Writing.swift
levigroker: I'm happy to change this... I don't understand the rationale, however. Are we concerned about collisions in the temp directory namespace?

Sources/ZIPFoundation/FileManager+ZIP.swift
levigroker: I understand the intention, but, from the documentation:

"Attempting to predicate behavior based on the current state of the file system or a particular file on the file system is not recommended. Doing so can cause odd behavior or race conditions. It’s far better to attempt an operation (such as loading a file or creating a directory), check for errors, and handle those errors gracefully than it is to try to figure out ahead of time whether the operation will succeed."

Tests/ZIPFoundationTests/ZIPFoundationWritingTests.swift
levigroker: Nice to have a reply for sure... I'm not in love with the response, however.

Tests/ZIPFoundationTests/ZIPFoundationWritingTests.swift
levigroker: Without it the memory usage grows since we are in a tight loop.

Tests/ZIPFoundationTests/ZIPFoundationWritingTests.swift
levigroker: Thanks!

Sources/ZIPFoundation/Archive+Writing.swift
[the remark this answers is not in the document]
levigroker: `remove/move` will succeed when `replace` fails in the case when the source and destination URLs represent content on different volumes.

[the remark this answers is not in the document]
levigroker: Yes, I did expose it so it could be tested... `private` feels is a bit aggressive, but if you maintain it should be `private` I'll have to look into trying to test this through a calling method.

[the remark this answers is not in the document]
levigroker: I looked at this a bit and it would be difficult to attempt to test this method through the caller. Essentially, this is called when attempting to remove an entry from a compressed archive. A temp archive is created in a temporary directory. To test the cross-link situation the temporary archive would have to reside on a different volume than the archive being manipulated, however, the new temporary directory method has been explicitly changed to create the temporary directory on the same volume. We would have to manufacture a failover mechanism such that the temporary directory creation fell back to the `NSTemporaryDirectory()` usage, while placing the main archive on a separate volume. That sounds like a significant effort that I'm not sure I'm able to commit to, considering the alternative is to leave this method exposed for testing.

Sources/ZIPFoundation/Archive+Writing.swift
weichsel: I saw that you already implemented the change - Thank you. 
Using the same method for UUID generation will make it easier to catch all occurrences when refactoring unique path creation (which I plan to move into a convenience method at some point).

Tests/ZIPFoundationTests/ZIPFoundationWritingTests.swift
weichsel: I just looked into that. No I see your point: The memory grows to several gigabytes in this loop 😱. I expected it to grow to some megabytes. 
Screenshot 2019-05-14 at 22 48 59
Instruments shows that a private `NSFileManager` method (`_URLForReplacingItemAtURL`) is allocating 10000s of 1kb blocks for some reason. 
Will have to further investigate what is going on here. This feels wrong.

Sources/ZIPFoundation/Archive+Writing.swift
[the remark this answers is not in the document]
weichsel: OK. No worries. I will have a look (hopefully tomorrow) if we can refactor that a bit or otherwise we can keep those methods `internal` for now.

Sources/ZIPFoundation/FileManager+ZIP.swift
levigroker: Does this make sense? From the docs, the return value is "true if the directory was created, true if createIntermediates is set and the directory already exists" so the check is not needed since if the directory already exists there will be no error.

weichsel:
@levigroker Short follow-up on this PR. Just want to let you know that I haven't forgotten about it.  
I am currently drowning in other tasks, but I will continue the review at the end of the week.  
Sorry for the delay.

levigroker:
@weichsel No stress from me. Thanks for the update though. :)

Tests/ZIPFoundationTests/ZIPFoundationWritingTests.swift
weichsel: I looked further into this.  
`FileManager.url(for:)` is implemented since Swift 5.0 in corelibs Foundation. Post 5.0 - in the current `master` -  there also have been some refinements to make the method behave more like the Darwin implementation.  
So we have to find workarounds for the following limitations on non-Darwin paltforms: 
 - we can't use `autoreleasepool` 
 - Swift 4.0, 4.1 & 4.2 need fallbacks because `url(for:)` is missing (e.g. by always returning a URL based on `NSTemporaryDirectory()` from `uniqueTemporaryDirectory`)

levigroker: *  I've removed the `autoreleasepool`... it's only a test anyway.
* I've added a condition for non darwin systems for `url(for:)` compatibility.

levigroker: I'm not sure what you'd like to do about the listing failures. They are about file and method length, and that's a bigger task than I care to deal with.

weichsel:
@levigroker I just fixed the remaining swiftlint issues. Let me know if those commits look good and I'll merge this.

levigroker:
Thanks @weichsel, your changes look good to me.

levigroker:
👌

levigroker:
🎉

weichsel:
@levigroker Finally merged - Thanks for your contribution (and your patience) Levi!

levigroker:
Glad to!

Issue #135, opened 2019-08-20
Suggestion: Replace CRC32 with call to zlib for better performance?


MrMage:
**Is your feature request related to a problem? Please describe.**
Despite #40, calculating the CRC32 is still slow, especially in debug builds. Given that `zlib` [exposes a CRC32 method](https://rosettacode.org/wiki/CRC-32#Swift) and `zlib` ships with all Macs, would you be open to replacing the built-in CRC32 computation with a call to `zlib`'s `crc32` method?

**Describe the solution you'd like**
`Data.crc32` calls into `zlib` instead of implementing the CRC32 computation itself.

**Describe alternatives you've considered**
Keep things as is, accepting slow performance in debug builds.

weichsel:
Hi Daniel,

I am aware of the slow performance of the built-in CRC32 method 🙁 but I try to keep the framework as self-contained as possible.  
The `DEBUG` version of `crc32(checksum:)` is ~ an order of magnitude slower than the `RELEASE` version. (And it gets even worse when you increment the `SWIFT_VERSION` to 5 and use the latest toolchain in the Xcode 11 betas). 

I have no clue why that relatively simple piece of code is _so_ slow when compiled unoptimized.  
The Time Profiler output is also hard to interpret because it contains a lot of `<Unknown Address> `sections. 🤷‍♂️

![Screenshot 2019-08-21 at 22 45 16](https://user-images.githubusercontent.com/1577319/63467533-facd2480-c465-11e9-856c-da30ecdd6c86.png)

I assume the slow CRC checks are holding you back during debugging? 
Would it help if we implement #120 ? That would allow you to conditionally turn off checksumming during debugging.

MrMage:
Thank you for the elaboration! My app automatically creates a ZIP backup archive once per week, so if I could disable writing CRC32's to the ZIP archive in debug builds that would work. Not sure if #120 would also cover _writing_ archives, though, and whether omitting CRC32s is even allowed by the ZIP format.

FYI, when I tried to sample the ZIP process, it looked like the bounds checks were what slowed things down. It _might_ be faster to directly access the Data's contained bytes instead, see https://github.com/vapor/template-kit/blob/master/Sources/TemplateKit/Utilities/HTMLEscape.swift for an example. However, @weissi strongly recommends _not_ using `unsafe` for purposes like this; maybe he has a suggestion for speeding this up in debug builds without using `unsafe`.

MrMage:
[...]

PR #137, opened 2019-08-23
Slightly speed up CRC32 calculation.

MrMage:
Yields a ~40% speedup in Release and ~10% in Debug builds for me. @weichsel, please make sure to try the change in this PR yourself and see if you see the same improvements.

Improves on #135.


weichsel:
The changes look 👌. 

> ... please make sure to try the change in this PR yourself and see if you see the same improvements.

I noticed similar speedups. The full test suite now runs ~ 16s faster in Release mode on my Mac (Down to 170s from 186s). Thank you for your contribution Daniel!

PR #139, opened 2019-10-05
Optional argument for passing on existing files

abinop:
optional argument on unzipItem in order to not throw an exception when a duplicate file is extracted. This happens when the zip contains __MACOSX entries

Fixes #

# Changes proposed in this PR
* 
* 
* 

# Tests performed

# Open Issues

### DIFF
--- a/Sources/ZIPFoundation/FileManager+ZIP.swift
+++ b/Sources/ZIPFoundation/FileManager+ZIP.swift
@@ -12,7 +12,7 @@ import Foundation
 
 extension FileManager {
     typealias CentralDirectoryStructure = Entry.CentralDirectoryStructure
-
+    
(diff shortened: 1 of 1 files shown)


weichsel:
Thank you for your pull request.  

While existing files are probably the most likely use case for a "skip item" feature, there are some other scenarios where skipping certain items could be useful (e.g. just overwriting _newer_ files or ignoring files that start with a certain pattern - `.DS_Store` for example).  

So instead of adding an additional parameter, a more flexible approach would be to implement a delegation mechanism similar to [`FileManagerDelegate`](https://developer.apple.com/documentation/foundation/filemanagerdelegate). 
Before extracting an item, the `unzipItemAt:` method should send a `fileManager:shouldUnzipItemAtURL:` message to its delegate and only proceed if the delegate returns `true`.

PR #143, opened 2019-10-30
Memory archive v2

microtherion:
Sorry about the long radio silence! I finally got around to working on https://github.com/weichsel/ZIPFoundation/pull/78 again, and ended up with a complete redesign that massively simplifies the semantics of the user API.

# Changes proposed in this PR
* Replace the previous use of `fmemopen` and `open_memstream` with a new abstraction `MemoryFile` which is implemented with `funopen` on Apple platforms and `fopencookie` on Linux. 
* Add `Archive.init?(data: Data = Data(), accessMode mode: AccessMode, preferredEncoding: String.Encoding? = nil)` for in-memory processing of either an existing `Data` or a new `Data` to be created.
* Add `Archive.data() -> Data?` which for in-memory archives returns the resulting file.
* So for `.read` processing, you'd do:
```
let archive = Archive(data: myData, accessMode: .read)
// Usual extraction, as with a file based archive
```
* For `.write` and `.update` processing, there is one additional call:
```
let archive = Archive(accessMode: .write)
// Add entries to archive, as with a file based archive
if let myData = archive.data {
   //
}
```
* On an unrelated matter, I've made `Data.compress` and `Data.uncompress` public, because I needed this for a different project (Gzip compression in a web server).

# Tests performed
* Added unit tests for both the low level `MemoryFile` functionality and the in-memory `Archive` abstraction on top of it.
* Passes on both macOS and Linux

# Open Issues
* The one flaw is that on Linux, this requires passing an extra flag to the Swift compilation. In Swift 5.0 and later, this is possible through `Package.swift`, and I've added the flag there. On Swift < 5, I believe this functionality is not yet available in the Swift package manager, so you have to manually add `-Xcc -D_GNU_SOURCE=1` to the `swift` invocation. I don't know whether it's possible to add the flag to the Cocoapods build.


weichsel:
Wow, this looks great - thank you for this follow-up PR!  

I haven't tried it yet, but the API looks very concise and elegant. 
I will review & test this PR over the coming days and will also think about how we could solve the package description problem. 
As last resort, we could hide in-memory support for Swift versions <5. 

If I understand it correctly, `fopencookie` is only available on platforms that provide GNU extensions. Maybe we should also exclude all in-memory archive code paths from platforms that don't offer API to associate memory with IO streams.

microtherion:
I seem to have found a solution for the configuration problem: I've added a C based target for Linux that emulates `funopen` with `fopencookie`. With this change, tests now pass on Linux with Swift 4.0, 4.1, 4.2, 5.0, and 5.1, and should also pass with all these versions on macOS etc.

microtherion:
This builds on macOS locally, but fails CI for Apple platforms. Any idea why?

weichsel:
Today I started to look into your new approach and it works great so far. 
The API is really easy to use and very similar to the file-backed variant. 
One thing I ran into is, that I tried to use the return value of the `.data{}` closure after appending an entry to a in-memory, writable archive:
```swift
var tempURL = URL(fileURLWithPath: NSTemporaryDirectory())
tempURL.appendPathComponent("test")
tempURL.appendPathExtension("zip")
if let archive = Archive(accessMode: .create),
    let payload = "teststring".data(using: .utf8),
    let data = archive.data {
    try? archive.addEntry(with: "test.xyz", type: .file, uncompressedSize: UInt32(payload.count)) { (pos, length) -> Data in
        let range = Range<Int>(uncheckedBounds: (lower: pos, upper: length))
        return payload.subdata(in: range)
    }
    if let entry = archive["test.xyz"] {
        var data = Data()
        _ = try? archive.extract(entry) { data.append($0) }
        String(data: data, encoding: .utf8).map { print($0) }
    }
   // `data` here just contains the initial 22 bytes of an empty ZIP archive.
    dump(data)
   // To get the current in-memory contents, I have to re-access the `data` property of the archive.
    try? archive.data?.write(to: tempURL)
}
```

Regarding support of older Swift versions: Thank you for providing a shim that would allow us to support in-memory archives on Swift 4.x.  
I am still trying to figure out a path for the 1.0 release of the library. 
ZIP Foundation currently uses a single target Xcode project that defines architectures, deployment targets etc. for all Apple platforms. While this works, it is not really supported by Xcode and people occasionally report problems with signing, building or App Store submission.  
The xcodeproj is only used when projects integrate the library via Cartage `--no-build` which - from my experience - is a reasonable setup. (apart from the occasional issues)
I already heard similar issues when using CocoaPods. 
So for the 1.0 release I am currently leaning towards going SPM only because this abstracts away all the build settings and works on all platforms. This move would cut off all Xcode versions below 11 (at least for iOS development) and if Xcode 11 is a requirement, we could also require at least a Swift 5.0 toolchain.  

I saw that you bumped the Swift version in the Xcode project to 5.0. This will lead to build errors for people that integrate via Carthage and still use Xcode 10.x. (This is also the reason why the CI fails - it still uses an old Xcode version. I still need to update the CI configuration).

microtherion:
I figured out how to downgrade the Xcode project to Swift 4.2. Now all tests are passing except for SwiftLint and CodeCoverage. Any idea what could be misbehaving there?

The behavior you're seeing with `archive.data` is expected: The `data` property does not provide a data item that is updated as entries get added, but a snapshot of the current state. I might need to document that detail better.

weichsel:
Thanks for looking into the failing CI checks. 
If we want to keep the current set of supported language versions we'd have to (re)set the `SWIFT_VERSION` in the xcodeproj even lower - to `4.0`. But at this point I think we should get rid of Swift 4 support altogether. I will release one more 0.9.x version (based on the current `development`) and then move on with Swift 5.x only. Do you think this is too aggressive? But IMO it's really hard to keep the project well-tested for all platform ⊗ 2 major language versions.

The SwiftLint check is failing because of some code formatting issues (it's validating against the standard SwiftLint ruleset). I can take care of that - Would it be OK if I push directly onto your branch?  

CodeCoverage checks if the line coverage is still at 100% - I can also look into this.

microtherion:
I cannot claim any insights on what Swift releases should remain supported. I moved my own project to 5.0 a few months ago, so personally I don't need 4.x support. Maybe, as a compromise, you could drop 4.0 and 4.1 support and support 4.2/5.0/5.1?

Feel free to push directly to my branch. Whatever is most efficient to resolve the remaining CI issues.

weichsel:
I finally found some time to fix the swiftlint issues, but I ran into some problems with the changes of the xcodeproj that were introduced with `2def06718f539da60b6ec06451e344164c4b2aea`. (I think Xcode re-created large parts of the file when the `ZIPFoundationCompat` target was added. And it also messed-up some settings while doing so. eg. the script phases were omitted and some. deployment targets were re-set.). 
I also wrestled with the module map a bit because it contained an absolute path but then eventually  gave up and removed the commits that introduced the `funopen` emulation 😕. 
I saw that you already put a lot of effort into the workaround but I think it adds quite some configuration overhead that is hard to manage/maintain/test for all setups. So would it work for you if we move forward with making in-memory archives a >Swift 5 only feature?  

The conditional compilation blocks in `MemoryFile.swift` were missing `os(watchOS)`. Was this an oversight or is `funopen` not working on the Watch (compilation seems to work - haven't tested it yet though)?

The Code Coverage check should be relatively easy to fix.
Screenshot 2019-11-10 at 23 43 40
Most uncovered code paths should be easy to reach. I am just not sure yet how to trigger a failing `MemoryFile.open` during testing.

microtherion:
> I think Xcode re-created large parts of the file when the ZIPFoundationCompat target was added
I re-generated the project after adding `ZIPFoundationCompat` to `Package.swift`. I didn't realize it was manually maintained.

> would it work for you if we move forward with making in-memory archives a >Swift 5 only feature
That's fine with me.

> missing os(watchOS). Was this an oversight
That was an oversight on my part. I think I just blindly copied this from some other location.

Not sure who would want to deal with zip files on a watch, but if they do, it should be supported.

WowbaggersLiquidLunch:
Hi! I just want to pop in here and say thank you for this!

I have been looking for a way to zip/unzip data in memory without going through any file I/O. This (and only this) solved my problem.

If I may chime in on support for Swift versions, it might be safe to cut off support for versions < 5.0 on Apple platforms, if necessary. Swift 5 has ABI stability for Apple platforms, so I assume many apps have switched to Swift 5 for it.

weichsel:
@microtherion Getting this PR into the `development` branch was one of the last entries on my 2019 bucket list. I eventually managed to add the missing docs for the new methods and to add some more tests over the holidays. 

I think we have a small memory management bug in the `open` method of `MemoryFile` because of an unbalanced retain - but I will check if this is really an issue on `development`.

Sorry for the stuttering progress when landing this feature and thank you for your endurance.  
Happy 2020 🎊!

microtherion:
Thanks for sticking with this PR! It appears I underestimated just how much polish was needed to get this product ready.
As for the memory leak—if you're referring to `cookie`, that should be released in `closeStub`.

weichsel:
Ah - thanks. I didn't notice the release in `closeStub`.  

There was still one code path that leaked the cookie: If the `Archive(data: ...)` initializer failed because of an invalid or non-existent `endOfCentralDirectoryRecord`, it returned `nil` and never closed `archiveFile`. `closeStub` was not called in that case.

PR #149, opened 2020-01-02
Support swift 5.0 on cocoapods

bugix:
Support swift 5.0 on cocoapods

### DIFF
--- a/Sources/ZIPFoundation/Entry.swift
+++ b/Sources/ZIPFoundation/Entry.swift
@@ -11,7 +11,7 @@
 import Foundation
 import CoreFoundation
 
-/// A value that represents a file, a direcotry or a symbolic link within a ZIP `Archive`.
+/// A value that represents a file, a directory or a symbolic link within a ZIP `Archive`.
 ///
 /// You can retrieve instances of `Entry` from an `Archive` via subscripting or iteration.
 /// Entries are identified by their `path`.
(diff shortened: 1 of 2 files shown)


Sources/ZIPFoundation/Entry.swift
```diff
@@ -11,7 +11,7 @@
 import Foundation
 import CoreFoundation
 
-/// A value that represents a file, a direcotry or a symbolic link within a ZIP `Archive`.
+/// A value that represents a file, a directory or a symbolic link within a ZIP `Archive`.
```
weichsel: Good catch - Thanks!

ZIPFoundation.podspec
```diff
@@ -7,12 +7,12 @@ Pod::Spec.new do |s|
   s.social_media_url = 'http://twitter.com/weichsel'
   s.authors = { 'Thomas Zoechling' => 'thomas@peakstep.com' }
   s.source = { :git => 'https://github.com/weichsel/ZIPFoundation.git', :tag => s.version }
-  s.swift_version = '4.0'
+  s.swift_versions = ['4.0', '5.0']
```
weichsel: As we also have some code paths for `SWIFT_VERSION = 4.2`, we maybe should also include `4.2` here.  
(I assume CocoaPods just passed this value to the compiler.  
As far as I know `SWIFT_VERSION` accepts the following values: `4.0`, `4.2` and `5.0`.

bugix: Jup, let's add 4.2 as well

weichsel: Thank you!

PR #154, opened 2020-02-16
Memory safety fix

### LINKED ISSUE #155: Compression and Decompression can fail on Linux due to API misusage
# Summary
Pointers obtained via `withUnsafeMutableBytes` are stored and used outside of the scope of the method's closure. This is a misusage of the API as pointers are only valid during the lifetime of the closure.

microtherion:
Fixes #155

# Changes proposed in this PR 
In the Linux compression path, the code in a few places used a raw pointer obtained by `withUnsafeMutableByes` *outside* of the closure argument passed to that function. This is not a correct use of that API, and in fact, I've observed in some situation that compression failed because the data pointed to changed between the call to `withUnsafeMutableBytes` and the use.

The fix is simply to nest the code using the data inside the closure argument. This looks a bit uglier due to the deeper nesting, but no longer exhibits the problems I saw earlier.

# Tests performed
Ran unit tests, and used the modified code in my app.

### DIFF
--- a/Sources/ZIPFoundation/Data+Compression.swift
+++ b/Sources/ZIPFoundation/Data+Compression.swift
@@ -236,31 +236,33 @@ extension Data {
         repeat {
             let readSize = Swift.min((size - position), bufferSize)
             var inputChunk = try provider(position, readSize)
-            stream.avail_in = UInt32(inputChunk.count)
-            inputChunk.withUnsafeMutableBytes { (rawBufferPointer) in
-                if let baseAddress = rawBufferPointer.baseAddress, rawBufferPointer.count > 0 {
-                    let pointer = baseAddress.assumingMemoryBound(to: UInt8.self)
-                    stream.next_in = pointer
-                }
-            }
             zipCRC32 = inputChunk.crc32(checksum: zipCRC32)
-            flush = position + bufferSize >= size ? Z_FINISH : Z_NO_FLUSH
-            var outputChunk = Data(count: bufferSize)
-            repeat {
-                stream.avail_out = UInt32(bufferSize)
-                outputChunk.withUnsafeMutableBytes { (rawBufferPointer) in
-                    if let baseAddress = rawBufferPointer.baseAddress, rawBufferPointer.count > 0 {
-                        let pointer = baseAddress.assumingMemoryBound(to: UInt8.self)
-                        stream.next_out = pointer
-                    }
-                }
-                result = deflate(&stream, flush)
-                guard result >= Z_OK  else {
+            stream.avail_in = UInt32(inputChunk.count)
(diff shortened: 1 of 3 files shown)


microtherion:
Hmm… it looks like `Data.decode` suffers from the same problem.

weichsel:
Good catch - Thanks for looking into this!

> Hmm… it looks like Data.decode suffers from the same problem. 
I haven't investigated the details yet, but IIRC `decode` follows a similar pattern. 
So moving the usage of the unsafe pointer into the scope of the `withUnsafeMutableBytes` closure should also work there, right?

Currently the Linux tests fail for Swift versions < 5:
Looks like the introduced `corruptedData` error triggers the failure (but not for the Swift 5.0 codepath 🤷‍♂️).
`ZIPFoundationWritingTests.swift:156: error: ZIPFoundationTests.testCreateArchiveAddZeroSizeCompressedEntry : failed - Failed to add zero-size compressed entry to archive with error : corruptedData`

microtherion:
> So moving the usage of the unsafe pointer into the scope of the withUnsafeMutableBytes closure should also work there, right?
Yes, I believe so.
> Currently the Linux tests fail for Swift versions < 5
I'll take a look.

microtherion:
* Fixed the Linux issues (when fed zero length data, `withUnsafeBytes` returns `nil`).
* Applied the memory safety fix to `decode` (Only tested in unit tests, not in any app).
* Changed the `TooLargeUncompressedEntry` unit test to generate zero filled instead of random data, because it was so slow, and spent a ton of time in the random number generator. This test is supposed to error out anyway, and I felt that using random data for the other tests was sufficient. But this is an isolated commit and I can revert it if you prefer.

microtherion:
I can refactor to fix the linter failures (presumably by taking some of the loop bodies into an auxiliary function). But it looks to me as if the previous code might already have been >=40 lines. Let me know what you prefer.

weichsel:
Thank you very much for looking into this.  

> Fixed the Linux issues (when fed zero length data, withUnsafeBytes returns nil).

Good catch!

> Applied the memory safety fix to decode (Only tested in unit tests, not in any app).

Additionally to the tests, I wrote a basic CLI package that exercised the decoding code paths on Linux and also reviewed the code. The CLI app behaved as expected and code-wise it definitely looks more correct now with the `deflate` call nested in the memory closure. 👍

> Changed the TooLargeUncompressedEntry unit test to generate zero filled instead of random data, because it was so slow, and spent a ton of time in the random number generator. This test is supposed to error out anyway, and I felt that using random data for the other tests was sufficient. But this is an isolated commit and I can revert it if you prefer.

Good idea - I think randomness is not required for that test and this test method was by far the slowest in the whole test suite.

PR #162, opened 2020-03-26
Convert 'Provider' to a protocol

bouk:
This allows passing in a Data or UnsafeBufferPointer into addEntry, without having to specify your own Provider function.

# Changes proposed in this PR

* Converting Provider to a Protocol

# Tests performed

I added a test that uses a UnsafeBufferPointer from a String, showing that this is now super easy.

This is obviously backwards-compatibility breaking. What we could do is add additional overrides for addEntry that take in a protocol, and keep the old ones with uncompressedSize and the Provider function. Then we could write a wrapper that takes those two values and creates a protocol implementation out of it.

# Open Issues

There's probably some style issues, but I've never used swiftlint before so I don't know how to get it to format my code properly. I'll have to look into that. First I'd like just some feedback on what you think of this change and whether you find it worthwhile. I think it will make the project a lot more ergonomic to use, I think it immediately improves the internal code too.

- [ ] Fix linux code (haven't looked at it yet)

### DIFF
--- a/Sources/ZIPFoundation/Data+Compression.swift
+++ b/Sources/ZIPFoundation/Data+Compression.swift
@@ -16,14 +16,51 @@ public typealias CRC32 = UInt32
 /// - Parameters:
 ///   - data: A chunk of `Data` to consume.
 /// - Throws: Can throw to indicate errors during data consumption.
-public typealias Consumer = (_ data: Data) throws -> Void
+public typealias Consumer<D: ContiguousBytes> = (_ data: D) throws -> Void
 /// A custom handler that receives a position and a size that can be used to provide data from an arbitrary source.
 /// - Parameters:
 ///   - position: The current read position.
 ///   - size: The size of the chunk to provide.
 /// - Returns: A chunk of `Data`.
 /// - Throws: Can throw to indicate errors in the data source.
-public typealias Provider = (_ position: Int, _ size: Int) throws -> Data
+public protocol Provider {
+    associatedtype SubSequence: ContiguousBytes
+
+    var count: Int { get }
+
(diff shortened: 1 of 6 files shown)


Sources/ZIPFoundation/Data+Compression.swift
```diff
@@ -16,14 +16,51 @@ public typealias CRC32 = UInt32
...
+    }
+}
+extension Data: Provider {}
+extension Array: Provider where Self.Element == UInt8 {}
+extension ArraySlice: Provider where Self.Element == UInt8 {}
+extension ContiguousArray: Provider where Self.Element == UInt8 {}
+extension UnsafeBufferPointer: Provider where Self.Element == UInt8 {}
+extension UnsafeRawBufferPointer: Provider {}
+extension UnsafeMutableBufferPointer: Provider where Self.Element == UInt8 {}
+extension UnsafeMutableRawBufferPointer: Provider {}
```
bouk: These types can now be passed in when a Provider is expected, without any wrapping.

weichsel:
Hi bouk,
Thank you for your PR but I think I don't fully understand your changes yet.  
What's the advantage of a protocol based approach over reader/writer closures here? 

> This allows passing in a Data or UnsafeBufferPointer into addEntry, without having to specify your own Provider function.  

The `Consumer`/`Provider` based read/write methods are the lowest level methods to operate on ZIP files in the framework and the ability to provide custom logic before passing data to the archive is their main purpose. E.g. when receiving data over a network, a `Provider` closure can contain logic to skip specific parts before forwarding the data to `addEntry`.  
If I understand your changes correctly, this would no longer be possible.  

Can you explain me the use case you are solving with the PR?
 Maybe this would help me to better understand the proposed changes.

PR #169, opened 2020-04-20
Fix version being incorrectly checked in isZIP64

OCJvanDijk:
ZIPFoundation failed reading an archive because it thought the entry was ZIP64, while in reality the entry had its OS byte set in versionNeededToExtract.

# Changes proposed in this PR
* Check version against the lower byte only in isZIP64 check, because upper byte signifies OS compatibility.


OCJvanDijk:
If it were up to me I would probably split both the versionMadeBy and versionNeededToExtract fields into two UInt8 fields to avoid future confusion.

weichsel:
Thanks for the PR and sorry for the super late reply.  

> ZIPFoundation failed reading an archive because it thought the entry was ZIP64, while in reality the entry had its OS byte set in versionNeededToExtract.  

The [ZIP specification](https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT) defines the `versionNeededToExtract` field to be two bytes. Currently there is no ZIP version that exceeds the limit of 1 byte - but there might be in the future. 
If the app/library that created the archive wrote a value that used the second byte, the bug is in the writing library.

OCJvanDijk:
Hey,

It's been a while, but it looks like I had gotten section 4.4.2 and 4.4.3 mixed up.
I guess I should have opened an issue first before submitting a PR.

Anyway, thanks for reviewing!

OCJvanDijk:
Ok, it's been a while, but I've run into this again and I'm 95% sure that I was right after all. I'll open an issue this time with a more detailed explanation.

weichsel:
> Ok, it's been a while, but I've run into this again and I'm 95% sure that I was right after all. I'll open an issue this time with a more detailed explanation.

I am also pretty sure now that you are right here. Thanks for being so persistent and sorry for closing the PR too early. 
Re-opened and merged.

PR #188, opened 2020-09-25
Fix the critical performance issue of `crc32` function.

andyqee:
Fixes #
Fix the performance issue of `crc32` function.
The `crc32` function which implemented in Swift has a serious performance issue in debug build without compiler optimization. 
So it will affect the performance of compress and uncompress operation 

# Changes proposed in this PR
* I replaced the crc32 swift implementation with the C based implementation which included in the System Lib Zlib

#### Beta build [No compiler optimization]
Device : iPhone 11 Pro 
Text size 24.2 MB

|                     | compress  | uncompress |   
|---------------------|-----------|------------|
| Before Optimization | 13.067s   | 12.918s    |  
| After Optimization  | 0.348s    | 0.0949s    |
| Result  |  38x    | 143x    | 

#### Release build [optimize for speed [-O] ]
Device : iPhone 11 Pro 
Text size 24.2 MB

|                     | compress  | uncompress |   
|---------------------|-----------|------------|
| Before Optimization | 0.404s   | 0.165s    |  
| After Optimization  | 0.340s    | 0.0919 s    |
| Result  |  1.18x    | 1.795x    | 

#### Release build [optimize for size [-Osize] ]
Device : iPhone 11 Pro 
Text size 24.2 MB

|                     | compress  | uncompress |   
|---------------------|-----------|------------|
| Before Optimization | 0.409s   | 0.167s    |  
| After Optimization  | 0.332s    | 0.0908 s    |
| Result  |  1.23x    | 1.839x    | 
Heaviest Stack trace from Instrument 

Before optimization
Screen Shot 2020-09-25 at 17 19 00

After optimization
Screen Shot 2020-09-25 at 17 11 17

# Open Issues

- [ ] Can't build with Linux && Swift 4.0 environment

### DIFF
--- a/Sources/ZIPFoundation/Data+Compression.swift
+++ b/Sources/ZIPFoundation/Data+Compression.swift
@@ -9,6 +9,13 @@
 //
 
 import Foundation
+#if os(macOS) || os(iOS) || os(watchOS) || os(tvOS)
+import zlib
+
+#else
+import CZlib
+
+#endif
 
 /// An unsigned 32-Bit Integer representing a checksum.
 public typealias CRC32 = UInt32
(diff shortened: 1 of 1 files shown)


noxo:
would be nice to get this PR in, i'm seeing 300x performance drops with crc32 on.

noxo:
@andyqee Does this improve performance also in Release builds? I've noticed there is a ~300x performance drop in Debug builds, but then it Release mode haven't notice big difference (yet)? #135 #137

weichsel:
Thank you for the PR and the benchmarks @andyqee .

As @noxo mentioned, the abysmal CRC32 performance is a symptom of Debug builds and should be reasonable when building with optimizations. 

On Darwin platforms, ZIP Foundation currently is self-contained and intentionally does not rely on zlib (trading performance during debugging for less dependencies).  
I'd prefer to keep it that way and instead of importing zlib, add a note to the README that explains the Debug vs. Release difference. It's also possible the improve the debugging experience by using a combination of `#if DEBUG` & `skipCRC32`.

andyqee:
@noxo 
Thanks for you comments.
I tested with release build as you said the performance difference is not that big.

#### Release build [optimize for speed [-O] ]
Device : iPhone 11 Pro 
Text size 24.2 MB

|                     | compress  | uncompress |   
|---------------------|-----------|------------|
| Before Optimization | 0.404s   | 0.165s    |  
| After Optimization  | 0.340s    | 0.0919 s    |
| Result  |  1.18x    | 1.795x    | 

#### Release build [optimize for size [-Osize] ]
Device : iPhone 11 Pro 
Text size 24.2 MB

|                     | compress  | uncompress |   
|---------------------|-----------|------------|
| Before Optimization | 0.409s   | 0.167s    |  
| After Optimization  | 0.332s    | 0.0908 s    |
| Result  |  1.23x    | 1.839x    |

weichsel:
@andyqee Thank you for re-doing your performance comparison with a release build!  
While the zlib based approach is slightly faster, I think this doesn't justify the additional dependency. 

I will close this PR and update the README to make people more aware of the huge performance discrepancy between Debug and Release builds.

noxo:
havent seen any updates to README?

witekbobrowski:
I would also love to see this being mentioned in the README. We are using ZIPFoundation to zip user's data (including photos) for export and we did notice **huge** compress times in `DEBUG`. I am happy I finally landed here to learn about the real reason behind slow compression. Compressing in `RELEASE` is blazing fast.

PR #190, opened 2020-10-23
Fix consumer is never called when extract uncompressed 0 byte entry

mtgto:
Fixes #

# Changes proposed in this PR
* Fixes bug which closure `consumer` is never called while extracting 0 byte entry from uncompressed archive.

# Tests performed
* Add `testExtractUncompressedEmptyFile`
  * `testExtractUncompressedEmptyFile.zip` is copy from `testExtractUncompressedDataDescriptorArchive.zip`

* Using XCTestExpectation to verify consumer closure surely called at once.

# Open Issues
N/A


Tests/ZIPFoundationTests/ZIPFoundationReadingTests.swift
```diff
@@ -269,4 +269,22 @@ extension ZIPFoundationTests {
...
+
+    func testExtractUncompressedEmptyFile() {
+        let expectation = XCTestExpectation(description: "Called extract consumer")
+        let archive = self.archive(for: #function, mode: .read)
+        guard let entry = archive["empty.txt"] else { XCTFail("Failed to extract entry."); return }
+        let progress = archive.makeProgressForReading(entry)
+        do {
+            _ = try archive.extract(entry, progress: progress) { (data) in
+                // Progress does not finished because Progress assumes totalUnitCount > 0
+                // XCTAssertFalse(progress.isFinished)
```
mtgto, line 281: When extract 0byte entry, `Progress#isFinished` returns false.
I don't know it is good way or not...
https://developer.apple.com/documentation/foundation/progress

IMO, It is intuitive that Progress has 1/1 progress for extracting 0byte entry.

weichsel: This is a good point - we currently handle this edge case properly. 
We are using a 1/1 progress for entries of type `.directory` already and could also use the same approach for zero-byte files.  

Since this PR handles the bug where the completion handler is not called for zero-byte files, I removed the progress related code from the test case you provided. I think the progress mis-reporting should be handled in a dedicated PR.

weichsel:
Thanks for providing this PR - good catch!

I saw that you also filed a similar issue for _compressed_ entries: #191
I haven't been able to reproduce it yet, but from reading your description it doesn't look like it can be closed with this PR?

mtgto:
> I saw that you also filed a similar issue for compressed entries: #191
I haven't been able to reproduce it yet, but from reading your description it doesn't look like it can be closed with this PR?

Hmm. In my memory, method is different betweeon compressed and uncompressed entry... (It's the reason I created an another issue)
I'll check again and report to #191 .

PR #192, opened 2020-11-17
Fall back to suffix based entry type detection

### LINKED ISSUE #170: Please update this line to 'return isDirectory ? .directory : .file'
https://github.com/weichsel/ZIPFoundation/blob/ec32d62d412578542c0ffb7a6ce34d3e64b43b94/Sources/ZIPFoundation/Entry.swift#L150

GRiMe2D:
Fixes #170 

# Changes proposed in this PR
* Fall back to path suffix based entry type detection

# Tests performed
* I found a ZIP file where `ZIPFoundation` marks it as file while it is a directory. Originally it was an EPUB file, but due to copyright issues I've deleted entires using `zip -d -r 'zip file.zip' 'paths...'`.

  Added a fixture and test called `testReadingEntryTypes` which checks read entry list with expected ones. Test properly failed without a patch

Test case introduces a struct for better readability. Had to implement `Equatable` protocol by hand due to missing auto-synthesis on swift 4.0

# Zip file

Here is a problematic zip file [testReadingEntryTypes.zip](https://github.com/weichsel/ZIPFoundation/files/5552697/testReadingEntryTypes.zip)


weichsel:
Hi GRiMe2D,

Archives that don't specify their entry types aren't 💯 correct, but I think it makes sense here to be liberal when reading such archive.

Thanks for your contribution, great work!

GRiMe2D:
Hi @weichsel, thanks for accepting PR and merging to the main branch. Can we actually create a release now? Because, our app relies on this patch and I want to replace patched fork with this original repository. Thanks!

weichsel:
> Can we actually create a release now

Still need to perform some cleanups/tests but I will try to release a version in the upcoming days.

PR #193, opened 2020-12-17
new addEntry api for different archive name and file path

### LINKED ISSUE #122: Archive addEntry requires same filesystem structure
# Summary

The method `addEntry(with path: String, relativeTo baseURL: URL, compressionMethod: CompressionMethod = .none, bufferSize: UInt32 = defaultWriteChunkSize, progress: Progress? = nil) throws` does not allow the caller to build an archive entry which has a different directory structure than that of the filesystem.

Additionally, `addEntry(...` requires requires a real directory to exist on the filesystem when adding a directory archive entry.

# Steps to Reproduce

Imagine wanting to create an archive with an internal structure which differs from the structure present on the file system. This API does not allow for abstract entries to be created.

# Expected Results

The API should allow for entries to be placed inside the archive at arbitrary locations regardless of their place on the filesystem.

# Actual Results

One must arrange the filesystem to match the desired archive structure prior to utilizing this API.

# Suggestions

The API could be changed to decouple the archive path from the URL. Something like `addEntry(as path: String, from url: URL?, ...` which allows the caller to construct the resulting archive path in any desired manner, while optionally placing the content from the given URL at the path location in the archive. The optional from url would allow for a directory entry to be added without needing to reference the filesystem, since the contents of this directory on the filesystem are ignored (see #121 ).

roznet:
Fixes #122 

# Changes proposed in this PR
* New addEntry function with independent name and fileURL 

# Tests performed
All the unit tests passed

I renamed the exiting API and provided the previous one as a function of the new one as the reverse wasn't possible, but no functionality changed, as the existing addEntry trivially calls the new one.


.gitignore
```diff
@@ -65,3 +65,6 @@ fastlane/report.xml
 fastlane/Preview.html
 fastlane/screenshots
 fastlane/test_output
+
+# Swift Package Manager
+.swiftpm
```
weichsel, line 70: Good catch!

weichsel (approved):
Great API refinement - Thanks for the PR!

PR #199, opened 2021-01-25
Enable Module Stability

JWShroyer:
Modify Build Settings and Update Swift to enable Module Stability

# Changes proposed in this PR
* Enables Module Stability so the framework doesn't have to be rebuilt with compatible dependency managers if the only thing different is the swift toolchain version.

# Tests performed
As no code was changed, there was no need to add/remove/modify any tests. All tests were run, however, and succeeded.

As Module Stability was something added in Swift 5.1, this does upgrade the Xcode project to use Swift 5. I did not update the minimum iOS deployment target even though Xcode 12.3 was requesting that change as well.

### DIFF
--- a/ZIPFoundation.xcodeproj/project.pbxproj
+++ b/ZIPFoundation.xcodeproj/project.pbxproj
@@ -190,7 +190,7 @@
 		OBJ_1 /* Project object */ = {
 			isa = PBXProject;
 			attributes = {
-				LastUpgradeCheck = 1020;
+				LastUpgradeCheck = 1230;
 				TargetAttributes = {
 					"ZIPFoundation::ZIPFoundation" = {
 						LastSwiftMigration = 1020;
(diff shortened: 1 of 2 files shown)


JWShroyer:
I'm realizing now when opening this PR that there are swift package manager files which seem to support older versions of Swift. If it's more important to maintain the earlier versions of Swift instead of having module stability, please let me know.

weichsel:
Hi,

Thanks for your PR Joshua! 
> I'm realizing now when opening this PR that there are Swift Package Manager files which seem to support older versions of Swift.

We currently support CocoaPods, Carthage and SPM as dependency managers. 
The .xcodeproj is only used when integrating via Carthage (or when manually adding the project to a workspace). The Xcode project format does _not_ support definition of a _range_ of supported Swift versions and we therefore have to set the lowest, currently supported version (which is `4` ATM). Since we try to support the previous major version of Swift, we can't drop `4` yet.

JWShroyer:
@weichsel thanks for the heads up! To make sure I understand, do you mind telling me why it's important to support version 4 of Swift? Is it just for folks who want to use this project via Carthage or manually whose projects aren't updated to 5 minimum or is it something else? This is just so I learn something.

weichsel:
> To make sure I understand, do you mind telling me why it's important to support version 4 of Swift? Is it just for folks who want to use this project via Carthage or manually whose projects aren't updated to 5 minimum or is it something else?

Yes, that's the main reason. Since the language changes between Swift 4 and 5 are not as big as in previous Swift versions (e.g. between 2 and 3), I think it's doable to support -1 of the current major version.

JWShroyer:
Ok! Thank you for the explanation! Feel free to close this PR and we can revisit this when Swift 6 comes out maybe. 😄

PR #205, opened 2021-02-12
Add precondition for bufferSize parameter

### LINKED ISSUE #204: Infinite loop when extract is called with bufferSize <= 0
# Summary

When extract method is called with parameter bufferSize <= 0, function never returns.

# Steps to Reproduce

Call extract with bufferSize parameter 0.

# Expected Results

Method should throw or precondition should fire.

# Actual Results

Function never returns

# Regression & Version

# Related Link

valeriyvan:
Fixes #204 

# Changes proposed in this PR
Add precondition for bufferSize parameter 

# Tests performed

# Open Issues


Sources/ZIPFoundation/Archive.swift
```diff
@@ -81,6 +81,8 @@ public final class Archive: Sequence {
         case missingEndOfCentralDirectoryRecord
         /// Thrown when an extract, add or remove operation was canceled.
         case cancelledOperation
+        /// Thrown when an extract was called with zero or negative `bufferSize` parameter.
```
weichsel, line 84:
```suggestion
        /// Thrown when an extract operation was called with zero or negative `bufferSize` parameter.
```

weichsel (changes requested):
Thanks for adding the exceptions!
The CI currently fails because the 2 new error code paths are not covered.

weichsel (approved):
The exceptions for the zero-size edge case are a great overall improvement. 
Also thanks for adding the additional tests to cover the error code paths!

PR #206, opened 2021-02-14
Fix incorrect usage of MemoryLayout.size for MemoryLayout.stride

valeriyvan:
Fixes

Fix incorrect usage of `MemoryLayout.size` for `MemoryLayout.stride`.

[MemoryLayout](https://developer.apple.com/documentation/swift/memorylayout):

> Always use a multiple of a type’s stride instead of its size when allocating memory or accounting for the distance between instances in memory.

# Changes proposed in this PR

Fix incorrect usage of `MemoryLayout.size` for `MemoryLayout.stride`.

# Tests performed

# Open Issues

### DIFF
--- a/Sources/ZIPFoundation/Data+Compression.swift
+++ b/Sources/ZIPFoundation/Data+Compression.swift
@@ -94,13 +94,12 @@ extension Data {
         // The typecast is necessary on 32-bit platforms because of
         // https://bugs.swift.org/browse/SR-1774
         let mask = 0xffffffff as UInt32
-        let bufferSize = self.count/MemoryLayout<UInt8>.stride
         var result = checksum ^ mask
         #if swift(>=5.0)
         crcTable.withUnsafeBufferPointer { crcTablePointer in
             self.withUnsafeBytes { bufferPointer in
                 let bytePointer = bufferPointer.bindMemory(to: UInt8.self)
-                for bufferIndex in 0..<bufferSize {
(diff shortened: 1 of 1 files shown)


weichsel:
> Fix incorrect usage of MemoryLayout.size for MemoryLayout.stride.

Not sure if the usage of size is correct or incorrect here (IMO it's fine since no padding/alignment won't happen for basic types), but when reviewing this, I wondered about the purpose of the whole line altogether. 

Any result != 1 (for both, `.size` or `.stride`) for `UInt8` would be a real surprise here 🙃.
I suggest we simply use `self.count` instead. I did a bit of research how that `bufferSize` calculation came to be - couldn't find any reason. Was that way from the beginning 🤷‍♂️.

valeriyvan:
> > Fix incorrect usage of MemoryLayout.size for MemoryLayout.stride.
> 
> Not sure if the usage of size is correct or incorrect here (IMO it's fine since no padding/alignment won't happen for basic types), but when reviewing this, I wondered about the purpose of the whole line altogether.

I am sure usage is incorrect, but has no effect on code.
> 
> Any result != 1 (for both, `.size` or `.stride`) for `UInt8` would be a real surprise here 🙃.

As said, there's no effect on code.

> I suggest we simply use `self.count` instead. I did a bit of research how that `bufferSize` calculation came to be - couldn't find any reason. Was that way from the beginning 🤷‍♂️.

Ok.

PR #209, opened 2021-02-15
Speed up crc32 calculation in performance test ~4 times

valeriyvan:
Fixes #

# Changes proposed in this PR
* Replace for in range loop with while loop with index arithmetic which speeds up crc32 calculation in performance test ~4 times.
* some other minor changes.

# Tests performed

# Open Issues


weichsel:
> Replace for in range loop with while loop with index arithmetic which speeds up crc32 calculation in performance test ~4 times.

Thanks for looking into this - CRC calculation is a major performance bottleneck (esp. in `DEBUG` builds), so performance improvements are highly welcome in that area.   

I reviewed the code and it works great - but I wasn't able to reproduce the performance gains. 
I can't remember why we went with the range based approach over using pointer arithmetic back then.  
How did you measure the 4x gain?  
I used the unit test case that you provided with an earlier PR, switched the `SWIFT_VERSION` to 5 in the .xcodeproj (which is 4 per default, for backwards comp. reasons) and then ran the `testPerformanceCRC32` from within Xcode.  

When comparing the measure results of this branch with `development`, the metrics were roughly the same.

valeriyvan:
Apple MacBook Pro M1.

Before, ~13s:
Screenshot 2021-02-18 at 00 40 26

After, ~6s:
Screenshot 2021-02-18 at 00 47 31

Speed up ~2 times.

On Intel speed up was 4 times.

arthurdapaz:
This most be merged. Swift **ranging** is terrible for low level tasks. **While** is the Swift to-go control flow statement to ensure assembly language fast for-looping. The ARM binary file generated with a **Range** is dozens of lines bigger than the one that performs the same task using **While**. Also, the tests published above by @valeriyvan matched on my non-M1 macbook pro (late 2017)

PR #212, opened 2021-02-21
Add Objective-C support to extension of FileManager

valeriyvan:
Fixes #

Package is impossible to use in mix Swift-Objective-C projects without writing some kind of wrapper.
Making whole package Objective-C compatible seems to be a big rework.
But `(NS)FileManager` is Objective-C class and it sounds reasonable if methods of its extension will be available to be used from Objective-C.

This PR proposes minor changes needed to make extension of `FileManager` available for Objective-C.

# Changes proposed in this PR
Makes following methods of `FileManager` extension available from Objective-C
* `zipItem(at sourceURL:to:shouldKeepParent:compressionMethod:progress:)`
* `unzipItem(at:to:skipCRC32:progress:preferredEncoding:)`

# Tests performed
If in general this PR would be accepted, I will add tests for new added methods.

n/a

# Open Issues
n/a

### DIFF
--- a/Sources/ZIPFoundation/Archive.swift
+++ b/Sources/ZIPFoundation/Archive.swift
@@ -27,7 +27,7 @@ let dataDescriptorStructSignature = 0x08074b50
 let centralDirectoryStructSignature = 0x02014b50
 
 /// The compression method of an `Entry` in a ZIP `Archive`.
-public enum CompressionMethod: UInt16 {
+@objc public enum CompressionMethod: UInt16 {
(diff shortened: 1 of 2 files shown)


weichsel:
> If in general this PR would be accepted, I will add tests for new added methods.

I don't have plans to add Obj-C support to the framework. If people want to use ZIP Foundation from a mixed Obj-C/Swift codebase, they can wrap ZIP Foundation on the call site.

PR #215, opened 2021-02-23
Allow skipping checksum calculation when writing zip archives

LGriffioen:
Fixes performance being very slow with writing out large zip files (100 MB+)

# Changes proposed in this PR
* Add a parameter to skip calculating checksums when writing out zip archives, similar to the existing option for reading
* Parameter has a default values of false, so existing code will work the same without any API changes

# Tests performed
- All unit tests pass
- Zip files were written out, and then reopened successfully via macOS Archive Utility

# Open Issues

### DIFF
--- a/Sources/ZIPFoundation/Archive+Writing.swift
+++ b/Sources/ZIPFoundation/Archive+Writing.swift
@@ -96,6 +96,7 @@ extension Archive {
     public func addEntry(with path: String, type: Entry.EntryType, uncompressedSize: UInt32,
                          modificationDate: Date = Date(), permissions: UInt16? = nil,
                          compressionMethod: CompressionMethod = .none, bufferSize: UInt32 = defaultWriteChunkSize,
+                         skipCRC32: Bool = false,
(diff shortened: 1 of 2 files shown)


weichsel:
The CRC32 field is mandatory in both, local file headers and central directory entries. 
From the ZIP file format specification: 
> 4.1.5 Data integrity MUST be provided for each file using CRC32.  

Consumers of the archive that strictly adhere to the specification might deem the archive corrupt. Since we have no control over which software is used to read the archive, we can't skip writing the crc32 field.

> Fixes performance being very slow with writing out large zip files (100 MB+)

The slow performance is most likely related to `DEBUG` builds and shouldn't affect binaries built using the Release build configuration.

LGriffioen:
thanks for the information! I did some tests in release mode and confirmed that it's not an issue at all there.

weichsel:
> thanks for the information! I did some tests in release mode and confirmed that it's not an issue at all there.

Thanks for confirming - Good to hear that this issue is resolved now.

PR #220, opened 2021-04-07
Fix overflow of file count

waynezhang:
Fixes file count overflow.

ZIPFoundation crashes at https://github.com/weichsel/ZIPFoundation/blob/cf10bbff6ac3b873e97b36b9784c79866a051a8e/Sources/ZIPFoundation/Archive%2BWriting.swift#L354 if file count exceeds 65535 (UInt16.max).

# Changes proposed in this PR
* Add overflow check for `NumberOfEntriesOnDisk` and `NumberOfEntriesInCentralDirectory `

# Tests performed
An error is thrown instead of crash if file count exceeds 65535.

### DIFF
--- a/Sources/ZIPFoundation/Archive+Writing.swift
+++ b/Sources/ZIPFoundation/Archive+Writing.swift
@@ -351,9 +351,21 @@ extension Archive {
         let centralDirectoryDataLengthChange = operation.rawValue * (dataLength + CentralDirectoryStructure.size)
         var updatedSizeOfCentralDirectory = Int(record.sizeOfCentralDirectory)
         updatedSizeOfCentralDirectory += centralDirectoryDataLengthChange
-        let numberOfEntriesOnDisk = UInt16(Int(record.totalNumberOfEntriesOnDisk) + countChange)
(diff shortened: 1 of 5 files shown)


waynezhang:
Swiftlint fails at `Archive.swift`. Is that OK to alter the file length limit to some other larger value? I read `Archive.swift` and think it's better to keep it as is (or maybe extract `EndOfCentralDirectoryRecord` extension to some other file?).

weichsel:
> I read Archive.swift and think it's better to keep it as is (or maybe extract EndOfCentralDirectoryRecord extension to some other file?).

Would it work to move the `CompressionMethod` enum to `Data+Compression` or aren't that enough removed lines to get Archive <400?

waynezhang:
> > I read Archive.swift and think it's better to keep it as is (or maybe extract EndOfCentralDirectoryRecord extension to some other file?).
> 
> Would it work to move the `CompressionMethod` enum to `Data+Compression` or aren't that enough removed lines to get Archive <400?

Great idea! That would be enough. I've made the change.

weichsel:
> Great idea! That would be enough. I've made the change.

Thanks for the PR and for adding the additional test coverage 👍

PR #227, opened 2021-06-22
Use zlib.crc32 when available to improve performance of Data.crc32()

scier:
Use zlib.crc32 when available to improve performance of Data.crc32() by 3x in Release mode, 3000x in Debug mode

# Tests performed

Unit tests pass
Performance test comparison (on-device in Release mode: 3x speedup. Simulator in Debug mode: 3000x speedup)
Performed bitwise comparison of pure-swift path and zlib.crc32 path for random buffers; results identical.

Our project was previously able to utilize ZIPFoundation only by forcing the optimizer on via settings overrides applied in the Podfile; otherwise, in Debug mode, reading/writing ZIP archives in our use case took >10s rather than ~1.3s. As we switch to SwiftPM, this tweak is no longer possible. However, with this modification, both Debug and Release builds can execute the operations in ~1s.

# Open Issues


ondrej-hanslik-csob-cz:
Is there actually a platform where inflate/deflate algorithms are available and zlib is not there? I believe you should always use zlib for crc32. By the way, it would be nice to move it to a separate type instead of that nasty typealias `CRC32 = UInt32`

```
import Foundation
Import zlib

struct Crc32: Equatable {
    private(set) var value: uLong

    init() {
        value = 0
    }

    init(value: uLong) {
        self.value = value
    }

    init(value: UInt32) {
        self.value = uLong(value)
    }

    mutating func update(with data: Data) {
        value = data.withUnsafeBytes { bytes in
            return crc32(value, bytes.baseAddress!.assumingMemoryBound(to: UInt8.self), uInt(data.count))
        }
    }
}
```

Sources/ZIPFoundation/Data+Compression.swift
```diff
@@ -10,6 +10,10 @@
 
 import Foundation
 
+#if os(macOS) || os(iOS) || os(watchOS) || os(tvOS)
+import zlib
+#endif
```
valeriyvan, line 15: `#if canImport(zlib)`?

scier: Only reason I didn't do that was because I couldn't test on all platforms that might support zlib, so I wanted to be conservative. But if folks would prefer to go broad and include all platforms with zlib, then I'd be more than happy to change my PR

scier:
> Is there actually a platform where inflate/deflate algorithms are available and zlib is not there?

No, but -- the manifest (and Data+Compression.swift) seems to specifically depend on zlib on systems where Compression is available (i.e. Apple platforms), so I didn't want to introduce a dependency on it just for crc32 when there's a fallback available.

> I believe you should always use zlib for crc32.

That would certainly let us get rid of some code and a corner case, if we were willing to introduce an explicit zlib dependency even on Apple platforms.

> By the way, it would be nice to move it to a separate type instead of that nasty typealias `CRC32 = UInt32`

I agree, that would be cleaner; but that seems orthogonal to this change and would introduce a lot more little changes, so it feels like it should be a separate PR.

scier:
No worries about the delay -- thanks for integrating this, glad it was useful!

PR #230, opened 2021-08-05
ZIP64 Support

### LINKED ISSUE #118: ZIP64 Support
The original .ZIP format had a 4 GiB (2^32 bytes) limit, which is no longer relevant on 64 bit architectures and, since version 4.5 of the specification (currently at 6.3.6), has been supplanted by ZIP64, which increases the limits to 16EiB (2^64 bytes). 

ZIPFoundation should be able to handle archives larger than 4 GiB by adopting the ability to work with ZIP64 archives.

https://en.wikipedia.org/wiki/Zip_(file_format)#ZIP64
https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT

Ckitakishi:
Fixes #118 

# Changes proposed in this PR
* Add zip64 structures
  * `Archive.Zip64EndOfCentralDirectoryRecord`
  * `Archive.Zip64EndOfCentralDirectoryLocator`
  * `Entry.Zip64ExtendedInformation`
    * Using in both Local Header File and Central Directory, if uncompressed size or compressed size exceeds the limit, extended information should include both uncompressed and compressed file size fields. (at least in the Local header File)  
* Support writing zip64 file

# Tests performed
* All unit tests pass and coverage rate is 100%.
* Going to resolve **SwiftLint errors** after reviewer finished.
* In unit tests, the limit size of each field is actually a mock size.
  * Also tested archive files over `4GB` on macOS, zip64 files were written out, and they can be opened successfully by double-click.

## Note
1. Since this PR is large enough, the remaining tasks will be handled in the **#Next step**(See below). Some methods don't work very well enough currently, for this reason some code was commented out in the test cases.
2. Modified the github action settings temporarily, will revert this change when everything is done.
3. Code review workflow: https://github.com/weichsel/ZIPFoundation/issues/118#issuecomment-893362662

## Document
Zip specification: https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT

Overall .ZIP file format:
```
[local file header 1]
[file data 1]
 . 
.
 .
[local file header n]
[file data n]
[archive extra data record] 
[central directory header 1]
.
.
.
[central directory header n]
[zip64 end of central directory record]
[zip64 end of central directory locator] 
[end of central directory record]
```

## Others
@weichsel Please take a look when you have time, there's no need to hurry~ 
@billparousis I remember you commented on the zip64 issue, please take look if you still interested in it.

# Next step
* Unzip zip64 file and reading part.
  * Reading zip64 file
  * Fix entry extraction
  * Fix entry removal
  * DataDescriptor
  * ...


Sources/ZIPFoundation/Archive+Zip64.swift
```diff
@@ -0,0 +1,156 @@
...
+
+enum ExtraFieldHeaderID: UInt16 {
+    case zip64ExtendedInformation = 0x0001
+}
+
+extension Archive {
+    struct Zip64EndOfCentralDirectory {
+        let record: Zip64EndOfCentralDirectoryRecord
+        let locator: Zip64EndOfCentralDirectoryLocator
+    }
```
weichsel, line 26: nit: `ZIP64EndOfCentralDirectory`

Sources/ZIPFoundation/Archive+Zip64.swift
```diff
@@ -0,0 +1,156 @@
...
+        let zip64EOCDRecordSignature = UInt32(zip64EOCDRecordStructSignature)
+        let sizeOfZip64EndOfCentralDirectoryRecord: UInt
+        let versionMadeBy: UInt16
+        let versionNeededToExtract: UInt16
+        let numberOfDisk: UInt32
+        let numberOfDiskStart: UInt32
+        let totalNumberOfEntriesOnDisk: UInt
+        let totalNumberOfEntriesInCentralDirectory: UInt
+        let sizeOfCentralDirectory: Int
+        let offsetToStartOfCentralDirectory: Int
```
weichsel, line 38: type/size

Sources/ZIPFoundation/Archive+Zip64.swift
```diff
@@ -0,0 +1,156 @@
...
+
+extension Archive {
+    struct Zip64EndOfCentralDirectory {
+        let record: Zip64EndOfCentralDirectoryRecord
+        let locator: Zip64EndOfCentralDirectoryLocator
+    }
+
+    struct Zip64EndOfCentralDirectoryRecord: DataSerializable {
+        let zip64EOCDRecordSignature = UInt32(zip64EOCDRecordStructSignature)
+        let sizeOfZip64EndOfCentralDirectoryRecord: UInt
```
weichsel, line 30: type/size

Sources/ZIPFoundation/Archive+Zip64.swift
```diff
@@ -0,0 +1,156 @@
...
+        let sizeOfCentralDirectory: Int
+        let offsetToStartOfCentralDirectory: Int
+        let zip64ExtensibleDataSector: Data
+        static let size = 56
+    }
+
+    struct Zip64EndOfCentralDirectoryLocator: DataSerializable {
+        let zip64EOCDLocatorSignature = UInt32(zip64EOCDLocatorStructSignature)
+        let numberOfDiskWithZip64EOCDRecordStart: UInt32
+        let relativeOffsetOfZip64EOCDRecord: Int
```
weichsel, line 46: type/size

Sources/ZIPFoundation/Archive+Zip64.swift
```diff
@@ -0,0 +1,156 @@
...
+        self.sizeOfCentralDirectory = data.scanValue(start: 40)
+        self.offsetToStartOfCentralDirectory = data.scanValue(start: 48)
+        self.zip64ExtensibleDataSector = Data()
+    }
+
+    init(record: Archive.Zip64EndOfCentralDirectoryRecord,
+         numberOfEntriesOnDisk: UInt,
+         numberOfEntriesInCD: UInt,
+         sizeOfCentralDirectory: Int,
+         offsetToStartOfCD: Int) {
```
weichsel, line 100: type/size

Sources/ZIPFoundation/Archive+Zip64.swift
```diff
@@ -0,0 +1,156 @@
...
+        withUnsafePointer(to: &numberOfDiskStart, { data.append(UnsafeBufferPointer(start: $0, count: 1))})
+        withUnsafePointer(to: &totalNumberOfEntriesOnDisk, { data.append(UnsafeBufferPointer(start: $0, count: 1))})
+        withUnsafePointer(to: &totalNumberOfEntriesInCD, { data.append(UnsafeBufferPointer(start: $0, count: 1))})
+        withUnsafePointer(to: &sizeOfCD, { data.append(UnsafeBufferPointer(start: $0, count: 1))})
+        withUnsafePointer(to: &offsetToStartOfCD, { data.append(UnsafeBufferPointer(start: $0, count: 1))})
+        data.append(self.zip64ExtensibleDataSector)
+        return data
+    }
+
+    init?(data: Data, additionalDataProvider provider: (Int) throws -> Data) {
```
weichsel, line 79: type/size

Sources/ZIPFoundation/Archive+Zip64.swift
```diff
@@ -0,0 +1,156 @@
...
+        self.numberOfDiskStart = record.numberOfDiskStart
+        self.totalNumberOfEntriesOnDisk = numberOfEntriesOnDisk
+        self.totalNumberOfEntriesInCentralDirectory = numberOfEntriesInCD
+        self.sizeOfCentralDirectory = sizeOfCentralDirectory
+        self.offsetToStartOfCentralDirectory = offsetToStartOfCD
+        self.zip64ExtensibleDataSector = record.zip64ExtensibleDataSector
+    }
+}
+
+extension Archive.Zip64EndOfCentralDirectoryLocator {
```
weichsel, line 114: Nit: `ZIP64EndOfCentralDirectoryLocator`

Sources/ZIPFoundation/Archive+Zip64.swift
```diff
@@ -0,0 +1,156 @@
...
+        var totalNumberOfDisk = self.totalNumberOfDisk
+        var data = Data()
+        withUnsafePointer(to: &zip64EOCDLocatorSignature, { data.append(UnsafeBufferPointer(start: $0, count: 1))})
+        withUnsafePointer(to: &numberOfDiskWithZip64EOCD, { data.append(UnsafeBufferPointer(start: $0, count: 1))})
+        withUnsafePointer(to: &offsetOfZip64EOCDRecord, { data.append(UnsafeBufferPointer(start: $0, count: 1))})
+        withUnsafePointer(to: &totalNumberOfDisk, { data.append(UnsafeBufferPointer(start: $0, count: 1))})
+        return data
+    }
+
+    init?(data: Data, additionalDataProvider provider: (Int) throws -> Data) {
```
weichsel, line 128: type/size

Sources/ZIPFoundation/Archive+Zip64.swift
```diff
@@ -0,0 +1,156 @@
...
+        guard data.count == Archive.Zip64EndOfCentralDirectoryLocator.size else { return nil }
+        guard data.scanValue(start: 0) == zip64EOCDLocatorSignature else { return nil }
+        self.numberOfDiskWithZip64EOCDRecordStart = data.scanValue(start: 4)
+        self.relativeOffsetOfZip64EOCDRecord = data.scanValue(start: 8)
+        self.totalNumberOfDisk = data.scanValue(start: 16)
+    }
+
+    init(locator: Archive.Zip64EndOfCentralDirectoryLocator, offsetOfZip64EOCDRecord: Int) {
```
weichsel, line 136: type/size

Sources/ZIPFoundation/Archive.swift
```diff
@@ -11,16 +11,15 @@
 import Foundation
 
 /// The default chunk size when reading entry data from an archive.
-public let defaultReadChunkSize = UInt32(16*1024)
+public let defaultReadChunkSize = Int(16*1024)
```
weichsel, line 14: type/size

Sources/ZIPFoundation/Archive.swift
```diff
@@ -111,8 +108,22 @@ public final class Archive: Sequence {
...
+            ?? UInt(endOfCentralDirectoryRecord.totalNumberOfEntriesInCentralDirectory)
+    }
+    var sizeOfCentralDirectory: Int {
+        zip64EndOfCentralDirectory?.record.sizeOfCentralDirectory
+            ?? Int(endOfCentralDirectoryRecord.sizeOfCentralDirectory)
+    }
+    var offsetToStartOfCentralDirectory: Int {
+        zip64EndOfCentralDirectory?.record.offsetToStartOfCentralDirectory
+            ?? Int(endOfCentralDirectoryRecord.offsetToStartOfCentralDirectory)
+    }
```
weichsel, line 125: Good idea to provide ZIP/ZIP64 agnostic accessors here!
The type/size problem applies here as well, though. On 32-bit platforms, the `Int` will use the default word size there and counteract the ZIP64 benefits.

Sources/ZIPFoundation/Entry+Zip64.swift
```diff
@@ -0,0 +1,114 @@
...
+//
+
+import Foundation
+
+extension Entry {
+    enum EntryError: Error {
+        case invalidDataError
+    }
+
+    struct Zip64ExtendedInformation {
```
weichsel, line 18: Nit: `ZIP64ExtendedInformation`

Sources/ZIPFoundation/Entry+Zip64.swift
```diff
@@ -0,0 +1,114 @@
...
+            compressedSize = try value(of: .compressedSize)
+            relativeOffsetOfLocalHeader = try value(of: .relativeOffsetOfLocalHeader)
+            diskNumberStart = try value(of: .diskNumberStart)
+        } catch {
+            return nil
+        }
+    }
+
+    static func scanForZip64Field(in data: Data, fields: [Field]) -> Entry.Zip64ExtendedInformation? {
+        guard !data.isEmpty else { return nil }
```
weichsel, line 97:
```suggestion
        guard data.isEmpty == false else { return nil }
```

Sources/ZIPFoundation/Entry+Zip64.swift
```diff
@@ -0,0 +1,114 @@
...
+        var offset = 0
+        var headerID: UInt16
+        var dataSize: UInt16
+        let extraFieldLength = data.count
+        while offset < extraFieldLength - 4 {
+            headerID = data.scanValue(start: offset)
+            dataSize = data.scanValue(start: offset + 2)
+            let nextOffset = offset + 4 + Int(dataSize)
```
weichsel, line 105: Would be great if we could use something more self-documenting for those magic numbers here:
Either `MemoryLayout.size(...)` if applicable or constants with descriptive names.

Sources/ZIPFoundation/FileManager+ZIP.swift
```diff
@@ -223,15 +223,16 @@ extension FileManager {
         return modDate
     }
 
-    class func fileSizeForItem(at url: URL) throws -> UInt32 {
+    class func fileSizeForItem(at url: URL) throws -> Int {
```
weichsel, line 226: type/size

Sources/ZIPFoundation/FileManager+ZIP.swift
```diff
@@ -223,15 +223,16 @@ extension FileManager {
...
         let fileManager = FileManager()
         guard fileManager.itemExists(at: url) else {
             throw CocoaError(.fileReadNoSuchFile, userInfo: [NSFilePathErrorKey: url.path])
         }
         let entryFileSystemRepresentation = fileManager.fileSystemRepresentation(withPath: url.path)
         var fileStat = stat()
         lstat(entryFileSystemRepresentation, &fileStat)
-        return UInt32(fileStat.st_size)
+        // `st_size` is a signed int value
```
weichsel, line 234: 👍

Tests/ZIPFoundationTests/ZIPFoundationArchieveTests+Zip64.swift
```diff
@@ -0,0 +1,96 @@
...
+                                        0x01, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+                                        0x4c, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+                                        0x5a, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00]
+        let zip64EOCDRecord = Archive.Zip64EndOfCentralDirectoryRecord(data: Data(eocdRecordBytes),
+                                                                       additionalDataProvider: {_ -> Data in
+                                                                        return Data() })
+        XCTAssertNotNil(zip64EOCDRecord)
+    }
+
+    func testArchieveInvalidZip64EOCERecordConditions() {
```
weichsel, line 29: Thanks for also taking the time to cover the error codepaths!

Tests/ZIPFoundationTests/ZIPFoundationProgressTests.swift
```diff
@@ -180,5 +180,40 @@ extension ZIPFoundationTests {
...
+            } catch let error as Archive.ArchiveError {
+                XCTAssert(error == Archive.ArchiveError.cancelledOperation)
+            } catch {
+                XCTFail("Failed to add entry to uncompressed folder archive with error : \(error)")
+            }
+        }
+        self.wait(for: [cancel], timeout: 20.0)
+        zipQueue.sync {
+            XCTAssert(progress.fractionCompleted > 0.5)
+//            XCTAssert(archive.checkIntegrity())
```
weichsel, line 215: Is the integrity check failing?

Tests/ZIPFoundationTests/ZIPFoundationWritingTests+Zip64.swift
```diff
@@ -0,0 +1,316 @@
+//
+//  ZIPFoundationWritingTests+Zip64.swift
```
weichsel, line 2: Would rename the file to `+ZIP64`

Tests/ZIPFoundationTests/ZIPFoundationWritingTests+Zip64.swift
```diff
@@ -0,0 +1,316 @@
...
+        let size = 64 * 64 * 2
+        let data = Data.makeRandomData(size: size)
+        let entryName = ProcessInfo.processInfo.globallyUniqueString
+//        let currentLFHOffset = archive.offsetToStartOfCentralDirectory
+        do {
+            try archive.addFileEntry(with: entryName, size: size, data: data)
+        } catch {
+            XCTFail("Failed to add zip64 format entry to archive with error : \(error)"); return
+        }
+        // Please ignore this test for now, will be fixed later
```
weichsel, line 159: Can this be un-commented?

Tests/ZIPFoundationTests/ZIPFoundationWritingTests+Zip64.swift
```diff
@@ -0,0 +1,316 @@
...
+        XCTAssertLessThan(0, archive.zip64EndOfCentralDirectory?.record.sizeOfCentralDirectory ?? 0)
+    }
+
+    func testRemoveEntryFromArchiveWithZip64EOCD() {
+        /*
+         File structure:
+         testRemoveEntryFromArchiveWithZip64EOCD.zip/
+           ├─ data1.random (size: 64)
+           ├─ data2.random (size: 64 * 64)
+         */
```
weichsel, line 256: Great comment!
Nit: Would use `//` single line comment syntax for all lines. (Instead of `/* */`)

weichsel (changes requested):
First of all: This is great work!
Thanks for implementing this and for the clean additions and the great test coverage!

The only real issue I found was the mixed usage of `(U)Int32`/`(U)Int64`/`(U)Int` - Due to the adaptive behavior of `(U)Int`, it could wipe the ZIP64 benefits on 32-bit platforms IMO.

Tests/ZIPFoundationTests/ZIPFoundationProgressTests.swift
Ckitakishi: Check failed because fixing for `extract(..)` is still working in progress (read methods are called there).

Tests/ZIPFoundationTests/ZIPFoundationWritingTests+Zip64.swift
Ckitakishi: Currently the answer is no 😢 , `readChunk(..)` is called in `makeIterator(..)`, thus these tests will be fixed in next step.

Ckitakishi:
@weichsel Thank your for pointing out the mixed usage of (U)Int32/(U)Int64/(U)Int, I have fixed this, however, a special case is `bufferSize`, since `fwrite(..)` and `fread(..)` take the `Int` size as a parameter, changing the type of `bufferSize` to `Int64` would make things more difficult, so I decided to keep using `Int`.

One more thing, in Darwin the `off_t` is always 64 bit, no need to take care of it. But according to [GNU C Library Manual](https://www.gnu.org/software/libc/manual/pdf/libc.pdf), in glib it might be 32 bit on 32 bit platform, so I tried to modified the `shim.h` file to enable `64 bit`, but I’m not sure if it correct, please help me to take a look at this part~

Tests/ZIPFoundationTests/ZIPFoundationProgressTests.swift
weichsel: Ah, I see - thanks.

weichsel:
> Thank your for pointing out the mixed usage of (U)Int32/(U)Int64/(U)Int, I have fixed this, however, a special case is bufferSize, since fwrite(..) and fread(..) take the Int size as a parameter, changing the type of bufferSize to Int64 would make things more difficult, so I decided to keep using Int.

Thanks for the fixes - I haven't tested thoroughly yet, but from glancing over the PR everything looks 👍. I think we stick with `Int` for the `bufferSize`- buffers shouldn't be sized  >=2^32 anyways.

> One more thing, in Darwin the off_t is always 64 bit, no need to take care of it. But according to GNU C Library Manual, in glib it might be 32 bit on 32 bit platform, so I tried to modified the shim.h file to enable 64 bit, but I’m not sure if it correct, please help me to take a look at this part~

Thanks for the detailed investigation here. After my first review round I noticed that the `_LARGEFILE64_SOURCE` is toggled via Xcode project setting? (At least it's set to `0` there ATM). IMO that setting doesn't have any effect because all non-Darwin platforms use SPM. The xcodeproj is only used by Carthage.

Ckitakishi:
> buffers shouldn't be sized >=2^32 anyways.

Thanks for the advice! Did you mean buffer size should less than `UInt32` on 64-bit device either..? This topic is really interesting, would you mind telling me more about the reason? 

IMO, both `Int` and `Int32` look good to me, but I don't think `UInt32` is a good choice for us. For example, if `bufferSize` is `UInt32.max`, then the `chunk`'s count can be `UInt32.max`, in this case, we could only get an `Int32` number by `chunk.count` on 32-bit device.

https://github.com/weichsel/ZIPFoundation/blob/417427e13f128fcfb388af2bed88665242793825/Sources/ZIPFoundation/Data%2BSerialization.swift#L89-L102


> After my first review round I noticed that _LARGEFILE64_SOURCE is toggled via Xcode project setting? (At least it's set to 0 there ATM)

Ah, I'm sorry, I added `_LARGEFILE64_SOURCE ` to Xcode project setting just for test, I'll remove it!

Ckitakishi:
@weichsel 
Hi 👋 , I created a PR for unzip last week, please check it out when you have time ~
https://github.com/Ckitakishi/ZIPFoundation/pull/6

billparousis:
@Ckitakishi Thanks for the great effort here.  I had a chance to go over the PR a bit, but I feel more confident with @weichsel reviewing this code, since I'm not entirely familiar with the ZipFoundation core codebase.  My team is hoping to test this against some large zip files we were having issues with once this is finalized.

Ckitakishi:
>  I discovered some API breaking changes during testing

@weichsel Really a good point! As a framework user I usually prefer the deprecation warnings to build errors.
I think maybe we can wait a while (say, one year?) before removing the old APIs entirely.

weichsel:
@Ckitakishi I found one more issue with this PR. We currently use a mixture of `Int64` and `UInt64`. Was this on purpose? 
Because the [.ZIP File Format Specification](https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT) states that all field (unless otherwise noted) should be unsigned:
> 4.4.1.1  All fields unless otherwise noted are unsigned and stored
> in Intel low-byte:high-byte, low-word:high-word order.

By using signed fields, we are cutting the (theoretical) file size limit in half. 
If this wasn't done by design, we should convert all `Int64` fields to `UInt64`. 

Can you have a look - or should I give It a try?  

Apart from that, the implementation looks really robust - so far everything I threw against it worked flawlessly.

weichsel:
Hi @Ckitakishi ,
Thanks for the detailed explanation - If I understand correctly, we are between a rock (the ZIP specification) and a hard place (the file system API) here 😐.

The different file size limits between signed and unsigned size values are of theoretical nature at the moment (I doubt there are many files out there that use sizes > `Int64.max`). But we can try to be consistent where possible and cast at the last moment (= when calling system API).  
IMO the only real problematic scenario is when parsing bytes from archives that we didn't create where a size value exceeds `Int64.max`. Should we throw there?

Ckitakishi:
@weichsel Your understanding is completely correct!

> But we can try to be consistent where possible and cast at the last moment (= when calling system API).

Actually I've thought about same solution. But IMO, there are several drawbacks:
1. It's better to throw when size value exceeds `Int64.max`.
2. We should avoid user to create a zip file exceeding `Int64.max` since it cannot be unachieved (should also throw exception when `compressSize` exceeds `Int64.max`)
3. Some user may be confused as the interface is `UInt64` but in fact we don't really support archiving/unarchiving zip file whose size value exceeds `Int64.max`

However, I agree that complying with the specification might be what we should do. If you still think it is better to convert all `Int64` fields to `UInt64`, I can give a try!

weichsel:
@Ckitakishi , I just performed a test conversion of all spec-fields (+ dependent values) to `UInt64`. All tests pass and in general I mostly had to perform "widening" casts (`Int64` -> `UInt64`). 
There were 2 locations where we rely on signed values because we perform addition of potentially _negative_ values:
- In `writeEndOfCentralDirectory` when applying `cdDataLengthChange`
- In `scanForZIP64EndOfCentralDirectory` because `recordOffset` can be negative for new (=empty) archives

I am not familiar enough with the latter case, but I assume that it's valid for new archives to _not_ have a `ZIP64EndOfCentralDirectory` record. 

> However, I agree that complying with the specification might be what we should do. If you still think it is better to convert all Int64 fields to UInt64, I can give a try!

That'd be great. I think you are most familiar with the additions needed for ZIP64 support and the implications re. to signed-ness. In my experiment I simply changed the `Int64` fields in the new structs and fixed types in methods that used them. (Incl. parameters of calling methods). `Int64` remained for progress related code (`totalUnitCount...`) and casted values when calling system file API.  

> It's better to throw when size value exceeds Int64.max.
We should avoid user to create a zip file exceeding Int64.max since it cannot be unachieved (should also throw exception when compressSize exceeds Int64.max)
Some user may be confused as the interface is UInt64 but in fact we don't really support archiving/unarchiving zip file whose size value exceeds Int64.max

Maybe we can we ensure that public methods restrict users to work with `Int64`? (e.g. `addEntry` -> using `Int64` for `uncompressed` size)

Ckitakishi:
@weichsel 
Hi, I made a draft implementation for converting all fields to `unsigned` type: https://github.com/Ckitakishi/ZIPFoundation/pull/7/files

After actually trying it out, my mind changed a bit because converting all fields to unsigned types anyway didn't seem to do any good except for making the design closer to the `ZIP64` specification. 

1. I don't think marking fields as `unsigned` types fully satisfies specification, because we still need to use them as signed values in some places
2. Use a mixture of `Int64` and `UInt64` for one field (e.g. uncompressSize) is a bit weird.
   - `addEntry` -> using `Int64` for uncompressed size
   - `uncompressedSize` field is marked as `UInt64`
   - The max `uncompressedSize` is `Int64.max`
3. We have to throw exceptions when casting `UInt64` to `Int64` to ensure the robustness of ZIPFoundation (even if type casting is safe enough).

Therefore, I prefer to keep the current implementation and add some negative checks for public methods. (But it totally depends on you! 😆 )

weichsel:
@Ckitakishi Thank you for your follow-up PR. 
I reviewed and tested it - everything's working and looking good now.  
Thanks for implementing this feature (and indulging my nitpicking) 🙃.

Merged 🎉

Ckitakishi:
@weichsel Very glad to see this feature has been merged! 🎉
Working on ZIPFoundation is a great experience for me, I'll report back if there are any problems(hope there aren't 😂).

LGriffioen:
I just updated from a November 2021 commit to latest and tracked down an 8x performance regression in debug (have not checked release yet) when loading Archives that was introduced with this changeset. I have not yet bisected the branch to pinpoint what commit is the culprit. Thought I'd share my discovery in case anyone can think of something that could cause this before digging deeper.

LGriffioen:
actually a bisect might not work here since this was never rebased onto https://github.com/weichsel/ZIPFoundation/pull/227 so the whole branch history will be much slower than what's merged into main.

weichsel:
> I just updated from a November 2021 commit to latest 

Is this a public commit? Can you provide a link?
Maybe CRC32 calculation is turned off there?

The only other change re. CRC32 that comes to mind is this one: d25ff70398eafb3ebd7d81dfe7dfbedfb51c4e39 - But that should also _improve_ performance 🤔.

> ... and tracked down an 8x performance regression

How did you measure? Did you use the same Xcode/toolchain versions? 

If you can find commit introducing a regression, please open a new issue.

LGriffioen:
thanks for the reply!

the commit we were on was https://github.com/weichsel/ZIPFoundation/commit/21429c6ed3f8dd50cdffb2c1bb2456ec4e19a522 which is the merge commit on main preceding the zip64 merge. the measurement was made using an internal performance logging framework we have in our app and i compared the same branch of our code with the same xcode/macOS, the only change was which commit of ZipFoundation I was pointing to.

I will see if i can narrow down which commit it was and I will open an issue if I can figure that out.

LGriffioen:
found the commit that introduced it, will write up an issue soon https://github.com/weichsel/ZIPFoundation/commit/b3ca1e6f050a26b6232d70fe04937bf2a7c282ba

PR #255, opened 2022-10-02
Correct README regarding the provided uncompressedSize

vanniktech:
Copied this blindly initially and wondered why some bytes were cut off from my string.


weichsel:
Good catch - Thank you for submitting a PR!

PR #256, opened 2022-10-10
Add public `Entry.isCompressed` property

marcomasser:
Fixes #

# Changes proposed in this PR
* Add a `Bool` property to check if a given `Entry` is compressed or not.

# Tests performed
I created a new ZIP archive that contains one compressed and one uncompressed `Entry`. The unit tests check whether the new `isCompressed` property matches the expected value for that.

I initially thought about adding a `var compressionMethod: CompressionMethod` property but decided against it. The issue with that is that the `CompressionMethod` is an `enum` with a `rawValue` that is read from the local file header, which might represent an unknown compression method. In that case, other code `throws` an error (e.g. `Archive.extract(_:bufferSize:skipCRC32:progress:consumer:)` throws `ArchiveError.invalidCompressionMethod`). It is possible to have a throwing property and do the same thing there, but I don’t know how ergonomic that really is. Also, in my intended use case, I don’t care what the actual compression method is – I just want to know whether the data is compressed or not. Therefore, I took the approach I saw in other places in the code that simply compares the compression value raw value against `CompressionMethod.none.rawValue`.

# Open Issues
- [ ] Should I commit the code for creating the test ZIP archive?


Tests/ZIPFoundationTests/ZIPFoundationEntryTests.swift
```diff
@@ -201,4 +201,10 @@ extension ZIPFoundationTests {
         XCTAssertEqual(dataDescriptorWithoutSignature?.uncompressedSize, 10)
         XCTAssertEqual(dataDescriptorWithoutSignature?.compressedSize, 10)
     }
+
+    func testEntryIsCompressed() throws {
+        let archive = self.archive(for: #function, mode: .read)
+        XCTAssert(archive["compressed"]?.isCompressed == true)
+        XCTAssert(archive["uncompressed"]?.isCompressed == false)
+    }
```
marcomasser, line 209: The code for creating this test archive was:
```swift
var archiveURL = ZIPFoundationTests.resourceDirectoryURL
archiveURL.appendPathComponent(#function.replacingOccurrences(of: "()", with: ""))
archiveURL.appendPathExtension("zip")
let newArchive = Archive(url: archiveURL, accessMode: .create)!

let data = Data("This is just some test data that might be compressed or not".utf8)
try newArchive.addEntry(with: "uncompressed", type: .file, uncompressedSize: Int64(data.count), compressionMethod: .none, provider: { position, size in
    let position = Int(position)
    return data[position ..< (position + size)]
})
try newArchive.addEntry(with: "compressed", type: .file, uncompressedSize: Int64(data.count), compressionMethod: .deflate, provider: { position, size in
    let position = Int(position)
    return data[position ..< (position + size)]
})
```

weichsel: Looks good - thanks for providing a test case covering the addition!

PR #269, opened 2023-01-25
Make the `DataError` type `public`

### LINKED ISSUE #268: Cannot handle `DataError`s because it is only visible internally
Hi @weichsel 👋🏼 Thanks for this awesome library! I'd like to share with you a problem we have and discuss if we could find a good solution 🙂

**Is your feature request related to a problem? Please describe.**
I'm using the library to create zip files and handle the errors thrown by the library to provide feedback to the users. The library throws 3 types of errors: `ArchiveError`, `CocoaError` and `DataError`. The problem is `DataError` is not `public`, so I cannot check when such an error is thrown, for example, when the device does not have enough free space to write a chunk of data, the error `Data.DataError.unwriteableFile` is thrown, but I cannot know it from the application side. 😕

**Describe the solution you'd like**
The easier solution is just make `DataError` `public` as well. Then, I would be able to check when these errors are thrown.

**Describe alternatives you've considered**
Another solution is for the library to catch such errors and throw an `ArchiveError` instead.

Thanks!

Juanpe:
Fixes #268 

The goal of this PR is to make the `DataError` type `public` so that allowing the developers to handle this type of error in their apps.

### DIFF
--- a/Sources/ZIPFoundation/Data+Serialization.swift
+++ b/Sources/ZIPFoundation/Data+Serialization.swift
@@ -23,7 +23,7 @@ protocol DataSerializable {
 }
 
 extension Data {
-    enum DataError: Error {
+    public enum DataError: Error {
         case unreadableFile
         case unwritableFile
     }


weichsel:
Thanks Juanpe - makes sense to expose this type 👍

PR #271, opened 2023-02-07
Change library type to dynamic

mateusforgi:
Hi, I would like to add the possibility to import the library with dynamic link type. I have a XCFramework which has the ZIPFoundation as its dependency, however I am not able to link the ZIPFoundation because SPM imports it as static library so at runtime my app crashes because  dyld Symbol for ZIPFoundation was not found.

# Changes proposed in this PR
* Change library type to dynamic

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -20,8 +20,7 @@ let package = Package(
         .macOS(.v10_11), .iOS(.v9), .tvOS(.v9), .watchOS(.v2)
     ],
     products: [
-        .library(name: "ZIPFoundation", targets: ["ZIPFoundation"]),
-        .library(name: "ZIPFoundationDynamic", type: .dynamic, targets: ["ZIPFoundation"])
+        .library(name: "ZIPFoundation", type: .dynamic, targets: ["ZIPFoundation"]),
     ],
     targets: targets,
     swiftLanguageVersions: [.v4, .v4_2, .v5]
--- a/Package.swift
+++ b/Package.swift
@@ -20,7 +20,8 @@ let package = Package(
         .macOS(.v10_11), .iOS(.v9), .tvOS(.v9), .watchOS(.v2)
     ],
     products: [
-        .library(name: "ZIPFoundation", targets: ["ZIPFoundation"])
+        .library(name: "ZIPFoundation", targets: ["ZIPFoundation"]),
+        .library(name: "ZIPFoundationDynamic", type: .dynamic, targets: ["ZIPFoundation"])
     ],
     targets: targets,
     swiftLanguageVersions: [.v4, .v4_2, .v5]


mateusforgi:
@weichsel do you think we can just change the ZIPFoundatiom to dynamic like? 

`.library(name: "ZIPFoundatiom", type: .dynamic, targets: ["ZIPFoundation"])`

On apple documentation it says this:
` Leave this parameter so Swift Package Manager can choose between static or dynamic linking (recommended). `

But on the other part it says this: 
`A library’s product can be either statically or dynamically linked. It’s recommended that you don’t explicity declare the type of library, so Swift Package Manager can choose between static or dynamic linking based on the preference of the package’s consumer.`

It is kind of confusing as it recommends the dynamic linking but It also says to let the type as nil. 
The issue her is that if I have a XCFrameworks that depends on Library B and Library B depends on ZIPFoundation, ZIPFoundation has to also be linked dynamic and not just Library B.

mateusforgi:
@weichsel can you take a look on that?

On apple documentation it says this:
`Leave this parameter so Swift Package Manager can choose between static or dynamic linking (recommended).
`
But on the other part it says this:
`A library’s product can be either statically or dynamically linked. It’s recommended that you don’t explicity declare the type of library, so Swift Package Manager can choose between static or dynamic linking based on the preference of the package’s consumer.
`
It is kind of confusing as it recommends the dynamic linking but It also says to let the type as nil.
The issue her is that if I have a XCFrameworks that depends on Library B and Library B depends on ZIPFoundation, ZIPFoundation has to also be linked dynamic and not just Library B.

weichsel:
Hi Mateus,

We used to have a dynamic target in the package manifest in the past. This comes with its own set of problems and might add some confusion for package consumers. I'd rather stick with the "let Xcode decide" approach. I am aware of the problems that can arise from letting Xcode do its thing, but you might be able to solve that problems at the XCFramework level.

Have you seen this thread on the Swift forums?
https://forums.swift.org/t/packaging-static-library-in-spm-package-for-ios-executable/41245/3

mateusforgi:
Hi, Thomas. 

Yes, I have seen it. There is another topic [here](https://forums.swift.org/t/issue-with-third-party-dependencies-inside-a-xcframework-through-spm/41977) about XCFramework with third part dependencies. 

Regarding solving it on XCFramework level. Xcode does not give me the option to choose the type of linking. See 
[here](https://forums.swift.org/t/issue-with-third-party-dependencies-inside-a-xcframework-through-spm/41977/5) 

See also[ this comment](https://forums.swift.org/t/issue-with-third-party-dependencies-inside-a-xcframework-through-spm/41977/3) about possible solutions. For me the only think that worked was making dynamic IOS-DFU-Library and ZIPFoundation.

Package.swift
```diff
@@ -20,7 +20,7 @@ let package = Package(
         .macOS(.v10_11), .iOS(.v9), .tvOS(.v9), .watchOS(.v2)
     ],
     products: [
-        .library(name: "ZIPFoundation", targets: ["ZIPFoundation"])
+        .library(name: "ZIPFoundation", type: .dynamic, targets: ["ZIPFoundation"]),
```
balavor, line 23:
```swift
        .library(name: "ZIPFoundation", targets: ["ZIPFoundation"])
        .library(name: "ZIPFoundationDynamic", type: .dynamic, targets: ["ZIPFoundation"]),
```

PR #278, opened 2023-06-08
1. add last modify date support

enefry:
1. add last modify date support
2. add file path multi encode support


weichsel:
The modification date can be retrieved via `.modificationDate` key in the dictionary returned by `Entry.fileAttributes`. 
Since merging #343, the modification date is also set on non-Darwin platforms when unzipping.

PR #283, opened 2023-08-09
Fixes so repo builds with Xcode 15

JagCesar:
This PR Adds the os(xrOS) conditional check, without it the build fails as it falls into the #else block.

This is required to build with Xcode 15 beta 6

### DIFF
--- a/Package@swift-4.0.swift
+++ b/Package@swift-4.0.swift
@@ -1,7 +1,7 @@
 // swift-tools-version:4.0
 import PackageDescription
 
-#if os(macOS) || os(iOS) || os(watchOS) || os(tvOS)
+#if os(macOS) || os(iOS) || os(watchOS) || os(tvOS) || os(xrOS)
 let dependencies: [Package.Dependency] = []
 #else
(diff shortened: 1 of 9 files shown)


osy:
FYI os(xrOS) does not work since beta 4. It should be os(visionOS). See https://github.com/weichsel/ZIPFoundation/compare/development...osy:ZIPFoundation:development

JagCesar:
> FYI os(xrOS) does not work since beta 4. It should be os(visionOS). See [development...osy:ZIPFoundation:development](https://github.com/weichsel/ZIPFoundation/compare/development...osy:ZIPFoundation:development)

I don't think that's correct, I'm building one of our products using my fork and it builds, archives and uploads to App Store Connect without issues. Building on Xcode Cloud using Xcode beta 6. 🤷‍♂️

osy:
I’m on beta 5 so maybe it’s a recent change? Regardless Apple is sticking to visionOS as marketing and changed all their API headers to refer to visionOS. I don’t think people should refer to xrOS.

weichsel:
@JagCesar Thanks for the PR! 
I think @osy is right - Apple switched to `visionOS` for conditional compilation. Will merge and then change that on `development`.

PR #289, opened 2023-09-20
Privacy Manifest

### LINKED ISSUE #285: Privacy Manifest
**Is your feature request related to a problem? Please describe.**
Apple has announced the[ Privacy Manifest](https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_data_use_in_privacy_manifests), required by apps and SDKs.

```Third-party SDKs need to provide their own privacy manifest files that record the types of data they collect. Your app’s privacy manifest file doesn’t need to cover data collected by third-party SDKs that your app links to.```

**Describe the solution you'd like**
When are you guys planning to include it?

### PR DESCRIPTION
**Pull request by weichsel**
Fixes #285 

# Changes proposed in this PR
* Adds privacy manifest including tracking and API usage

### CODE REVIEW

File: Resources/PrivacyInfo.xcprivacy [hunk 1]
```diff
@@ -0,0 +1,32 @@
...
+            </dict>
+        </array>
+        <key>NSPrivacyAccessedAPITypes</key>
+        <array>
+            <dict>
+                <key>NSPrivacyAccessedAPIType</key>
+                <string>NSPrivacyAccessedAPICategoryFileTimestamp</string>
+                <key>NSPrivacyAccessedAPITypeReasons</key>
+                <array>
+                    <string>3B52.1</string>
```
zhenma: @weichsel From a code search, ZIPFoundation calls stat() (a required reason API) in FileManager+ZIP.swift. However, it does not call UIDocumentPickerViewController or UIDocumentBrowserViewController, and it does not reference UIKit anywhere, so it looks like ZIPFoundation does not specifically request users to grant access to timestamps of files or directories.

Users may grant access to access timestamps of files or directories in an app that integrates with ZIPFoundation, but that would be the app's responsibility to declare 3B52.1, instead of ZIPFoundation.

Could you elaborate why do you declare 3B52.1 for ZIPFoundation? Or would you consider removing it?

weichsel:
As pointed out in [this comment](https://github.com/weichsel/ZIPFoundation/issues/313#issuecomment-2011786133), `0A2A.1` seems to be a more appropriate API type reason. `3B52.1` will be replaced by `0A2A.1` in the next release.

PR #299, opened 2023-10-19
(fix) provides correct file url in thrown error with permission error

mredig:
Fixes #
Provides the file with permissions lacking, not the url of the zip archive, in the error thrown.

# Changes proposed in this PR
* 
* 
* 

# Tests performed
* all of them (Ran tests after change)

The current behavior is to throw an error when adding an entry to a zip file that cannot be read. This is perfectly reasonable, but the error thrown gives the url of the zip file, not the file with a permission error. The fix simply provides the newly added (or attempted, more like) file url, not the url of the zip file.

# Open Issues


weichsel:
Good catch - Thanks for the PR!

Issue #313, opened 2024-03-01
The collection purpose for the "Privacy Nutrition Label Types" is absent in the Privacy Manifests.


ika020202:
# Summary
The collection purpose for the "Privacy Nutrition Label Types" is absent in the Privacy Manifests.

If Privacy Nutrition Label Types are not needed, the item0 declared must be removed.
Otherwise, the appropriate The collection purpose should be declared.

Now, the privacy report is output as shown in the following image.
<img width="1637" alt="ZIPFoundation_PrivacyReport" src="https://github.com/weichsel/ZIPFoundation/assets/42883378/3330e525-585b-4c5b-b345-b81e95910089">

# Related Link
Data Use:
https://developer.apple.com/app-store/app-privacy-details/#:~:text=types%20not%20mentioned-,Data%20use,-You%20should%20have

For more information, please refer to the following thread and other library implementations.
ref: https://github.com/CocoaPods/CocoaPods/issues/10325

iPHYZLL:
+1 here. 

For now, I've added a blank entry to the `xcprivacy` on my side, but is this something that can be fixed in one of the upcoming releases?

if app is not using any of the `NSPrivacyCollectionDataTypes`, we can have an empty key in the `xcprivacy` to avoid the `Missing an expected key: 'NSPrivacyCollectedDataTypes'` message the generated `.pdf` file.

Empty key:
```
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<array/>
</plist>
```

dimayurkovski:
hello!
i think bug in podspec file, xcprivacy should be in s.resource_bundles

`s.resource_bundles = {'ZIPFoundation' => ['Sources/ZIPFoundation/Resources/PrivacyInfo.xcprivacy']}`

pietermuller:
I think the reason at `NSPrivacyAccessedAPICategoryFileTimestamp` should be `0A2A.1` (instead of `3B52.1`) as defined at:

Describing use of required reason API | Apple Developer Documentation
https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api
> 0A2A.1
[...]

PR #314, opened 2024-03-14
Fix privacy manifest issues

Westacular:
Fixes two issues related to the privacy manifest:

1. The manifest itself is kind of invalid (as discussed in #313)

Including a NSPrivacyCollectedDataTypes entry but specifying empty strings is incorrect. Instead, you can just specify an empty array for NSPrivacyCollectedDataTypes at the top level. Also, the top level NSPrivacyTracking and NSPrivacyTrackingDomains keys were missing; I've added them with false/empty values as appropriate.

2. The podspec reference for the privacy manifest was problematic.

When static framework linking is used with Cocoapods, files referenced using `resource` are just directly copied into the main app bundle, potentially clobbering each other or main app resources in the process. i.e., the PrivacyInfo.xcprivacy from ZIPFoundation can overwrite the one for the app itself.

The correct way to include PrivacyInfo.xcprivacy is by using `resource_bundles` to create a nested bundle to contain the file. (If dynamic linking is used, this bundle will be nested inside the framework bundle for the pod; if static linking is used, the bundle is nested directly in the main app bundle. In either case, clobbering is avoided.)

(Reference: https://github.com/CocoaPods/CocoaPods/issues/10325)

### DIFF
--- a/Sources/ZIPFoundation/Resources/PrivacyInfo.xcprivacy
+++ b/Sources/ZIPFoundation/Resources/PrivacyInfo.xcprivacy
@@ -2,21 +2,12 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
     <dict>
+        <key>NSPrivacyTracking</key>
+        <false/>
         <key>NSPrivacyCollectedDataTypes</key>
-        <array>
-            <dict>
-                <key>NSPrivacyCollectedDataType</key>
-                <string></string>
-                <key>NSPrivacyCollectedDataTypeLinked</key>
-                <false/>
-                <key>NSPrivacyCollectedDataTypeTracking</key>
-                <false/>
-                <key>NSPrivacyCollectedDataTypePurposes</key>
-                <array>
-                    <string></string>
-                </array>
-            </dict>
-        </array>
+        <array/>
+        <key>NSPrivacyTrackingDomains</key>
+        <array/>
         <key>NSPrivacyAccessedAPITypes</key>
         <array>
             <dict>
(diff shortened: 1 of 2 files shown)


weichsel:
Hi, 
Thanks for providing this PR. The fixes for the podspec look great 👍. 
I just tested the proposed changes to the privacy manifest itself and ran into a few issues though:

> Including a NSPrivacyCollectedDataTypes entry but specifying empty strings is incorrect. Instead, you can just specify an empty array for NSPrivacyCollectedDataTypes at the top level

I just pointed the dependencies of a large project I am involved with to your branch and noticed that the Xcode privacy report no longer contains any entries for ZIP Foundation. 
When reverting back to the current release tag, the privacy report properly contains an entry for ZIP Foundation (integrated via Xcode SPM):
Screenshot 2024-03-15 at 09 36 45

Westacular:
I can't find any explicit confirmation of this, but I believe that the purpose of those Xcode privacy reports is _specifically_ to help app developers fill out the "Privacy Nutrition Label" forms, with regard to what data is collected, how, and why.

i.e., it's _only_ surfacing details within `NSPrivacyCollectedDataTypes`. The otherwise empty line with the dash at the bottom of your screenshot is where the report should list the purposes (from `NSPrivacyCollectedDataTypePurposes`) for the data collection. ZIPFoundation isn't specifying a valid value there, and Xcode's showing a broken-looking empty string as a result.

The details in `NSPrivacyAccessedAPITypes` aren't surfaced in the Xcode privacy report at all, and frameworks that don't collect data and don't list anything in `NSPrivacyCollectedDataTypes` are ignored by the Xcode report, but it's my understanding (although the docs aren't clear) that privacy manifests that only have details in `NSPrivacyAccessedAPITypes` (and none in `NSPrivacyCollectedDataTypes`) are _not_ ignored by App Store Connect when analyzing an uploaded app... but I just don't know because Apple's docs and tools aren't clear on this. 🙁

RobertManeaSST:
Hi, any chance of this getting merged soon? I currently can't run my app if I create a Privacy file for it as well, as I'm getting the Multiple commands produce error for the PrivacyInfo from ZIPFoundation

PR #322, opened 2024-06-07
Fixed: Swift error in file Archive+MemoryFile.swift

### LINKED ISSUE #320: Project won't build on Ubuntu 24.04 with Swift 5.10 Release
# Summary

Compilation error encountered when running `swift build` on Ubuntu 24.04 with swift 5.10 Release

# Steps to Reproduce

1. Set up the project environment on Ubuntu 24.04.
2. Run the command `swift build`.

# Expected Results

The project should compile successfully without any errors.

# Actual Results

The following error is thrown during the compilation:

In file : `/ZIPFoundation/Sources/ZIPFoundation/Archive+MemoryFile.swift:45:20:`
```
error: value of optional type 'UnsafeMutablePointer<FILE>?' (aka 'Optional<UnsafeMutablePointer<_IO_FILE>>') must be unwrapped to a value of type 'UnsafeMutablePointer<FILE>' (aka 'UnsafeMutablePointer<_IO_FILE>')
fseeko(result, 0, SEEK_END)
```

# Small fix I have been using so far

I modified the MemoryFile() function slightly, just so the error does not get thrown during building.

```
class MemoryFile {
    private(set) var data: Data
    private var offset = 0

    init(data: Data = Data()) {
        self.data = data
    }

    func open(mode: String) -> FILEPointer? {
        let cookie = Unmanaged.passRetained(self)
        let writable = mode.count > 0 && (mode.first! != "r" || mode.last! == "+")
        let append = mode.count > 0 && mode.first! == "a"
        #if os(macOS) || os(iOS) || os(tvOS) || os(visionOS) || os(watchOS) || os(Android)
        let result = writable
            ? funopen(cookie.toOpaque(), readStub, writeStub, seekStub, closeStub)
            : funopen(cookie.toOpaque(), readStub, nil, seekStub, closeStub)
        #else
        let stubs = cookie_io_functions_t(read: readStub, write: writeStub, seek: seekStub, close: closeStub)
        let result = fopencookie(cookie.toOpaque(), mode, stubs)
        #endif
        if append {
            if let unwrappedResult = result {
                fseeko(unwrappedResult, 0, SEEK_END)
            } else {
                // Handle the error if result is nil
                print("Error: result is nil when attempting to open in append mode")
                return nil
            }
        }
        return result
    }
}

```

Dirk-Vantas:
Fixes #320

Fixed this smal error occuring during building of the package : 
```
error: value of optional type 'UnsafeMutablePointer<FILE>?' (aka 'Optional<UnsafeMutablePointer<_IO_FILE>>') must be unwrapped to a value of type 'UnsafeMutablePointer<FILE>' (aka 'UnsafeMutablePointer<_IO_FILE>')
fseeko(result, 0, SEEK_END)
```
# Changes proposed in this PR
The proposed change in the MemoryFile class aims to address a build error related to the unwrapping of an optional value. The error message indicated that result, which is of the type UnsafeMutablePointer<FILE>? (an optional pointer to a file), must be explicitly unwrapped before it can be used in the fseeko function call.

### DIFF
--- a/Sources/ZIPFoundation/Archive+MemoryFile.swift
+++ b/Sources/ZIPFoundation/Archive+MemoryFile.swift
@@ -42,11 +42,16 @@ class MemoryFile {
         let result = fopencookie(cookie.toOpaque(), mode, stubs)
         #endif
         if append {
-            fseeko(result, 0, SEEK_END)
+            if let unwrappedResult = result {
+                fseeko(unwrappedResult, 0, SEEK_END)
+            } else {
+                return nil
+            }
         }
         return result
     }
 }
+
 
 private extension MemoryFile {
     func readData(buffer: UnsafeMutableRawBufferPointer) -> Int {


weichsel:
Hi Dirk,

Thanks for making me aware of this. I just merge #323, which obsoletes `append` mode for the `MemoryFile` type. 
Can you please check `development` to see if this fixes your build issue?

Dirk-Vantas:
Hey Thomas

the dev branch builds without problems :)

Dirk-Vantas:
The project I am working on uses the main branch tho, when do you think this will be merged into that?

PR #331, opened 2024-08-26
Fix a possible memory leak in Data.readChunk

kbinani:
There is a memory leak in `Data.readChunk` when `fread` fails to read data.
This PR fixes it by deallocating memory before throwing an exception.

I didn't create an issue for this, as the problem is obvious, but please let me know if an issue is needed.


weichsel:
Thanks - Good catch!

PR #332, opened 2024-08-28
Fix CI Failure on Linux by Running Swift Tests Directly with Docker

kbinani:
When submitting another PR, I noticed that the CI on Linux was failing. It seems that instead of specifying `container:` in the CI.yml, we can resolve this issue by directly running the `swift test` command using docker.

I'm not sure what the GitHub Actions runner does when the `container:` option is specified, but there doesn't appear to be a significant difference in CI execution time between using `container:` and directly running the docker command.

- **Using `container:`:** 3m52s ~ 5m18s (example: [here](https://github.com/weichsel/ZIPFoundation/actions/runs/9446769361))
- **Using Docker command:** 3m49s ~ 6m18s (example: [here](https://github.com/kbinani/ZIPFoundation/actions/runs/10587729515))

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -34,11 +34,9 @@ jobs:
       matrix:
         tag: ['5.1', '5.2', '5.3', '5.4', '5.5', '5.6', '5.7', '5.8', '5.9', '5.10']
     runs-on: ubuntu-latest
-    container:
-      image: swift:${{ matrix.tag }}
     steps:
       - uses: actions/checkout@v2
-      - run: swift test -c release -Xswiftc -enable-testing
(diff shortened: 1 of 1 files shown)


weichsel:
Hi @kbinani 
Thanks for making me aware of this. I resolved the issue by removing the older Swift versions from the Linux test matrix. 
The problem arose because GitHub now defaults to `node20`, and the Swift images for versions 5.1-5.6 rely on a glibc version that's incompatible with the newer Node versions. Since testing those older Swift versions doesn't provide significant value, I opted to stick with the `container`/`image` support provided by GitHub Actions and simply removed the outdated images from the test matrix.

Issue #109, opened 2019-04-09
API to remove multiple entries at once


Rico99:
Currently entries can be removed one at a time, each time a temporary archive is created that replaces the original.

Would you consider extending the API to allow removing a list of entries?

For example:

    public func remove(_ entries: [Entry], bufferSize: UInt32 = defaultReadChunkSize, 
                                      progress: Progress? = nil) throws

I'm currently looking into adding this myself, but I have not wrapped my head around `CentralDirectoryStructure` and `endOfCentralDirectoryRecord` and the various offsets that are involved.

Thank you!

weichsel:
Removing entries from ZIP archives is - in contrast to entry addition - very expensive. So providing a "batch removal" method would certainly make sense.  
ZIP Foundation removes an entry by re-writing the whole archive to a temp file, but skipping the entry to remove. The part to skip is from `Entry.centraldDirectoryStructure.relativeOffsetToLocalHeader` to the end of the data section of the file. Wikipedia has a good introduction + diagram about the structure of archives:  
https://en.wikipedia.org/wiki/Zip_(file_format)#Structure

PR #333, opened 2024-08-29
New function to efficiently remove the last entries from an archive

depth42:
# Changes proposed in this PR
In this pull request, I added a new function for incrementally removing the last entries from an archive. This is very handy, when you want to remove and re-add entries add the end of an archive as you don't have to rewrite the archive in full.
The function is: Archive.removeAllEntries(fromEntry entry: Entry)

# Tests performed
I implemented trhee tests for the function following the existing test pattern.

- testRemoveFromEntryUncompressed
- testRemoveFromEntryCompressed
- testRemoveEntriesFromArchiveWithZIP64EOCD


weichsel:
Thanks for providing this PR. I like the idea of using truncate as fast-path when possible but I wonder if it wouldn't be better to keep that an implementation detail.
 We could check wether the entry to remove is the last one and use the code in the PR as alternate but private code path. (and provide a generic [remove multiple entries at once](https://github.com/weichsel/ZIPFoundation/issues/109) API as follow-up)

PR #337, opened 2024-09-06
fix ~10x performance regression by reverting a change

### LINKED ISSUE #311: Performance regression when opening lots of zip archives
# Summary
We have a ton of uncompressed zip archives that are read and written using ZipFoundation, and users can load all of them at once in a library view. I have 8,000+ items and before the commit linked below, these could all load in 1.2s, but after it takes over 12 seconds in debug. In release mode it was 1s before and 6s after.

Both tests were done on an M1 Max machine

# Steps to Reproduce
1. create thousands of uncompressed zips, can contain multiple files
2. load them all in a loop

# Expected Results
should be very fast, even in debug

# Actual Results
it's gotten ~6x slower in release, and 10x slower in debug

# Regression & Version
introduced in https://github.com/weichsel/ZIPFoundation/commit/b3ca1e6f050a26b6232d70fe04937bf2a7c282ba

# Related Link

LGriffioen:
change was introduced with zip64 support

Fixes #311 

# Changes proposed in this PR
* reverted https://github.com/weichsel/ZIPFoundation/commit/b3ca1e6f050a26b6232d70fe04937bf2a7c282ba#diff-d829f12a7b0fa38c616a3397c9ea1464287b0d0097dee344876fff88b4b28e39L316

# Tests performed
- performance of loading 10,000 files that are zipped is ~10x faster than development, and roughly where it was before the Zip64 changes
- I'm not familiar enough with Zip64 to be able to test what impacts this change has on that functionality

- Performance with large numbers of files is critical for us. This 10x improvement is taking a 20s operation down to under 2s

# Open Issues

### DIFF
--- a/Sources/ZIPFoundation/Archive.swift
+++ b/Sources/ZIPFoundation/Archive.swift
@@ -21,6 +21,7 @@ public let defaultDirectoryPermissions = UInt16(0o755)
 let defaultPOSIXBufferSize = defaultReadChunkSize
 let defaultDirectoryUnitCount = Int64(1)
 let minEndOfCentralDirectoryOffset = Int64(22)
+let maxDirectoryEndOffset = Int64(66000)
 let endOfCentralDirectoryStructSignature = 0x06054b50
 let localFileHeaderStructSignature = 0x04034b50
 let dataDescriptorStructSignature = 0x08074b50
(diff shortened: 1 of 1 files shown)


LGriffioen:
@weichsel @Ckitakishi i just rebased this, any chance either of you could take a look? thanks!

weichsel:
Thanks for keeping this up to date. 
I checked the changes, and they look good. They follow the ZIP specification, and since they only affect parsing of the standard EOCD record, there shouldn’t be any side effects or regressions related to ZIP64 support. This should also address the overall EOCD parsing performance regression. 🙏

LGriffioen:
thanks for taking a look!

PR #339, opened 2024-09-30
Fix alignment crash in debug mode

microtherion:
Fixes #

Recently when testing a nightly Swift snapshot in a project of mine, I came across an alignment crash. It turns out that

1. There is no guarantee that `Data` or `subdata` point to memory with any particular alignment.
2. In this configuration, with `T` being `UInt32`, the `subdata` was in fact *inline* data, which is always aligned on a 2 byte boundary but misaligned for 4 bytes.
3. When compiling in debug mode, `load` checks for proper alignment and aborts in this case.

The solution is to use the `loadUnaligned` call which makes no alignment assumptions.

# Tests performed

All unit tests keep passing.

Unfortunately, I have no good isolated reproduction case for the crash, nor can I be more specific about the project in question.

# Open Issues

### DIFF
--- a/Sources/ZIPFoundation/Data+Serialization.swift
+++ b/Sources/ZIPFoundation/Data+Serialization.swift
@@ -29,12 +29,9 @@ extension Data {
     }
 
     func scanValue<T>(start: Int) -> T {
-        let subdata = self.subdata(in: start..<start+MemoryLayout<T>.size)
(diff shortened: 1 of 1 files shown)


weichsel:
Thanks! Very good catch - I also wasn't aware of `loadUnaligned`.

microtherion:
Thank you!

Issue #356, opened 2025-03-30
Linux compilation error for main on `swift:latest` Docker image


dagronf:
# Summary

I have a Linux-based project which I'm trying to add Zip support.  When I add ZipFoundation as a dependency, the build fails with a compiler error. I'm using the `swift:latest` docker image on  Ubuntu 24.04.

However, if I change to using the `swift:5.8.1` docker image, the ZipFoundation library compiles successfully.

`swift:latest` version information

```
root@411ecbcc19f9:/src# swift --version
Swift version 6.0.3 (swift-6.0.3-RELEASE)
Target: x86_64-unknown-linux-gnu
```

```
root@dc032e84b02f:/src# cat /etc/os-release
PRETTY_NAME="Ubuntu 24.04.1 LTS"
NAME="Ubuntu"
VERSION_ID="24.04"
VERSION="24.04.1 LTS (Noble Numbat)"
VERSION_CODENAME=noble
ID=ubuntu
ID_LIKE=debian
```

When I add the ZIPFoundation package and try to compile, I receive a compiler error message.

```
root@411ecbcc19f9:/src# swift build
warning: 'zipfoundation': found 1 file(s) which are unhandled; explicitly declare them as resources or exclude from the target
    /src/.build/checkouts/ZIPFoundation/Sources/ZIPFoundation/Resources/PrivacyInfo.xcprivacy
Building for debugging...
/src/.build/checkouts/ZIPFoundation/Sources/ZIPFoundation/Archive+MemoryFile.swift:45:20: error: value of optional type 'UnsafeMutablePointer<FILE>?' (aka 'Optional<UnsafeMutablePointer<_IO_FILE>>') must be unwrapped to a value of type 'UnsafeMutablePointer<FILE>' (aka 'UnsafeMutablePointer<_IO_FILE>')
 43 |         #endif
 44 |         if append {
 45 |             fseeko(result, 0, SEEK_END)
    |                    |- error: value of optional type 'UnsafeMutablePointer<FILE>?' (aka 'Optional<UnsafeMutablePointer<_IO_FILE>>') must be unwrapped to a value of type 'UnsafeMutablePointer<FILE>' (aka 'UnsafeMutablePointer<_IO_FILE>')
    |                    |- note: coalesce using '??' to provide a default when the optional value contains 'nil'
    |                    `- note: force-unwrap using '!' to abort execution if the optional value contains 'nil'
 46 |         }
 47 |         return result
[...]

PR #358, opened 2025-06-16
Fix missing optional check on result in Ubuntu Noble

agentk:
Fixes https://github.com/weichsel/ZIPFoundation/issues/356

# Changes proposed in this PR

* Add an optional check for result in `MemoryFile.open(mode:)`

The issue is not really with result but with the API change to `fseeko` in Ubuntu Noble as it now requires the first parameter to be non-optional.


weichsel:
@ agentk  Thanks - the issue is already fixed on `development`. Will close this and tag a release later.

PR #361, opened 2025-07-21
Manually decode Code page 437 strings

### LINKED ISSUE #341: Legacy encoded zip archives won't unzip on Linux.
# Summary
When the UTF-8 flag in a zip archive is omitted, `ZIPFoundation` is trying to use CP437. This works on Apple platforms, but is not supported on Linux. As a result, all filenames default to `""`, and unzipping fails.

If Wikipedia is to be believed, CP437 in the spec is more theoretical, and in practice, legacy archives used whatever code page was set on the system. So maybe the easiest solution might be to pick an arbitrary 8-bit ASCII extension that *is* supported on Linux (e.g. I used `.latin1` successfully for my purposes).

# Related Link
Reported the problem on the Swift side: https://github.com/swiftlang/swift-foundation/issues/1029. Maybe they'll get around to fixing this (they seemed interested when I mentioned that in principle, CP437 is mandated by the ZIP spec):

davedelong:
Fixes #341 

# Changes proposed in this PR
* Implements a manual CP437 → UTF8 conversion, based on the wikipedia page

# Tests performed
* All unit tests pass


davedelong:
Perhaps, but I think that's beyond the scope of this PR. The encoding selection logic matches what was there before. My changes are simply to work around the fact that Foundation on Linux has no support for decoding CP437 data.

weichsel:
@davedelong Thanks for the PR - merged. I made your manual decoding codepath Linux-only: https://github.com/weichsel/ZIPFoundation/pull/367

PR #369, opened 2025-10-09
Improved symlinks support

F0xWh1sp3r:
There are archives containing compressed symlinks, and you'd like to be able to unzip them as well. Additionally, some symlinks may point to files in other subfolders within the archive.

# Changes proposed in this PR
* Allow extracting archives where symlinks are compressed
* Support symlinks pointing to files in other subfolders within the archive

# Tests performed
* Implemented three new tests: testUnzipSymlink, testUnzipCompressedSymlink, testInvalidSymlinkCompressionMethodErrorConditions
* All unit tests pass


weichsel:
Thanks for providing this PR. Changes look good!

Some things I noticed:
- The `FileManager` extension doesn't expose the new `allowUncontainedSymlinks`.
- The `allowUncontainedSymlinks` parameter is a bit redundant now. Passing a URL pointing to the FS root via `symlinksValidWithin` has the same effect right? Should we deprecate `allowUncontainedSymlinks`?
- There's a Swiftlint issue in one of the tests files.

F0xWh1sp3r:
Thank you for your time and review!

- Added new `symlinksValidWithin` to the `FileManager ` extension.
- You are right that passing a URL pointing to the FS root via `symlinksValidWithin` has the same effect. Yes, we should deprecate allowUncontainedSymlinks. With the `symlinksValidWithin` parameter, we can specify the required boundaries for symlinks. I marked the old functions as `deprecated` .

weichsel:
Great - Thanks for the changes!

PR #375, opened 2026-01-09
Add info-zip unicode path extra field

### LINKED ISSUE #373: Cannot extract entry whose path is encoded by GB 18030

tang3w:
Fixes #373 

# Changes proposed in this PR
* This PR adds Info-ZIP unicode path extra field when parsing local header, to facilitate ZIPFoundation to unzip the entries that don't have utf-8 encoded path in central directory structure but in Info-ZIP extra field.

# Tests performed
Tested with the test.zip attached in #373 

N/A

# Open Issues
N/A


Tests/ZIPFoundationTests/ZipFoundationEntryTests+InfoZip.swift
```diff
@@ -0,0 +1,35 @@
...
+        XCTAssertEqual(infoZip?.headerID, ExtraFieldHeaderID.infoZipUnicodePath.rawValue)
+        XCTAssertNotNil(infoZip)
+    }
+
+    func testInfoZIPUnicodePath() {
+        let fileManager = FileManager()
+        let archive = self.archive(for: #function, mode: .read)
+        let destinationURL = self.createDirectory(for: #function)
+        XCTAssertNoThrow(try fileManager.unzipItem(at: archive.url, to: destinationURL))
+    }
```
weichsel: Thanks for adding coverage for the additional codepaths!

ZIPFoundation.xcodeproj/project.pbxproj
```diff
@@ -15,6 +15,8 @@
 		59248AA226B6EC770078ABDA /* ZIPFoundationErrorConditionTests+ZIP64.swift in Sources */ = {isa = PBXBuildFile; fileRef = 59248AA126B6EC770078ABDA /* ZIPFoundationErrorConditionTests+ZIP64.swift */; };
 		5968EBED26D692360037EA62 /* ZIPFoundationReadingTests+ZIP64.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5968EBEC26D692360037EA62 /* ZIPFoundationReadingTests+ZIP64.swift */; };
 		5968EBF126D695AD0037EA62 /* ZIPFoundationFileManagerTests+ZIP64.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5968EBF026D695AD0037EA62 /* ZIPFoundationFileManagerTests+ZIP64.swift */; };
+		82BD5DB42F1112F300DA5764 /* Entry+InfoZip.swift in Sources */ = {isa = PBXBuildFile; fileRef = 82BD5DB32F1112D900DA5764 /* Entry+InfoZip.swift */; };
```
weichsel: Nit: Casing of the filename is off. Should be `Entry+InfoZIP`.

Tests/ZIPFoundationTests/ZIPFoundationTests.swift
```diff
@@ -296,6 +296,13 @@ extension ZIPFoundationTests {
         ]
     }
 
+    static var infoZipTests: [(String, (ZIPFoundationTests) -> () throws -> Void)] {
+        return [
+            ("testEntryScanForInfoZip", testEntryScanForInfoZip),
+            ("testInfoZIPUnicodePath", testInfoZIPUnicodePath)
+        ]
+    }
```
weichsel: Nit: Those tests can be merged into the common `zip64Tests`.

Sources/ZIPFoundation/Archive+ZIP64.swift
```diff
@@ -15,6 +15,7 @@ let zip64EOCDLocatorStructSignature = 0x07064b50
 
 enum ExtraFieldHeaderID: UInt16 {
     case zip64ExtendedInformation = 0x0001
+    case infoZipUnicodePath = 0x7075
```
weichsel: nit: The casing for `infoZip` should be `infoZIP` all the way through. 
Since `ExtraFieldHeaderID` is no longer solely about ZIP64, we could move it into `Archive.swift` (e.g. right below the other enums `Archive.swift:118`)

Sources/ZIPFoundation/Entry+Serialization.swift
```diff
@@ -65,6 +65,8 @@ extension Entry.LocalFileHeader {
         if let zip64ExtendedInformation = Entry.ZIP64ExtendedInformation.scanForZIP64Field(in: self.extraFieldData,
                                                                                            fields: self.validFields) {
             self.extraFields = [zip64ExtendedInformation]
+        } else if let infoZipUnicodePath = Entry.InfoZipUnicodePath.scanForUnicodePath(in: self.extraFieldData) {
+            self.extraFields = [infoZipUnicodePath]
         }
```
weichsel, line 69: A large ZIP64 archive could probably also have Unicode paths, so we should combine the fields. This will also indicate that those extra data fields aren't mutually exclusive if we add more of them in the future.

Sources/ZIPFoundation/Entry.swift
```diff
@@ -119,6 +119,10 @@ public struct Entry: Equatable {
     }
     /// The `path` of the receiver within a ZIP `Archive`.
     public var path: String {
+        // Read Info-ZIP Unicode Path extra field if present
+        if let infoZipExtraField = self.infoZipExtraField {
+            return String(pathData: infoZipExtraField.unicodeName, encoding: .utf8)
```
weichsel: As per spec, we should verify the CRC32 of the original filename here. We have a `crc32` helper in `Data+Compression`.

weichsel (changes requested):
Thank you for adding support for the Info-ZIP path extra field! This way we can decode paths in Info-ZIP compliant archives in a deterministic way on all platforms 🎉.

ZIPFoundation.xcodeproj/project.pbxproj
tang3w: Renamed the file

Sources/ZIPFoundation/Archive+ZIP64.swift
tang3w: Makes sense, moved the enum to the suggested place

Sources/ZIPFoundation/Entry+Serialization.swift
tang3w: Updated

Sources/ZIPFoundation/Entry.swift
tang3w: Added the CRC32 verification

weichsel:
Thank you for the changes - great work!

PR #380, opened 2026-05-04
Fix Android compile: explicit Bionic imports + Bionic-strict pointer signatures

odrobnik:
## Summary

Lets ZIPFoundation compile on the Swift Android SDK without changing behaviour on any other platform.

`import Foundation` does not transitively re-export Bionic the way it re-exports Darwin / Glibc / ucrt, so every file that touches a raw libc symbol fails on Android with errors like:

```
error: cannot find 'S_IFMT' in scope
error: cannot find 'stat' in scope
error: cannot find 'lstat' in scope
error: cannot find 'fopen' in scope
…
```

Three changes:

1. **Add an Android/Bionic import shim** at the top of every source file that uses libc symbols (`stat`, `lstat`, `S_IFMT`/`S_IFREG`/`S_IFDIR`/`S_IFLNK`, `mode_t`, `fopen`/`fclose`/`fread`/`fwrite`/`fseeko`/`ftello`, `fileno`, `ftruncate`, `time_t`/`gmtime`/`timegm`, `timeval`/`timespec`, `fpos_t`, `funopen`, `FILE`, `errno`):

   ```swift
   #if canImport(Android)
   import Android
   #elseif canImport(Bionic)
   import Bionic
   #endif
   ```

   Inert on Apple (Darwin via Foundation), Linux (Glibc via Foundation), and Windows (ucrt via Foundation) — Android is the only platform that gains new bindings.

2. **Wrap `setSymlinkPermissions` / `setSymlinkModificationDate` in their existing Apple-only `#if`.** Both are already only *called* on Apple (Linux's `lchmod` is a kernel no-op for symlinks; Bionic ships neither `lchmod` nor a useful `lutimes` for symlinks). Lifting the same `#if os(macOS) || …` over the *definitions* lets Android compile without changing Apple semantics.

3. **Fix two strict-signature mismatches** that surface only on Bionic:
   - `Archive+BackingConfiguration.swift`: `fwrite(buffer.baseAddress, …)` — Bionic types the first parameter as non-optional `UnsafeRawPointer`. Adopt the same `if let baseAddress, count > 0` pattern already used in `Data+Serialization.write(chunk:)`.
   - `Archive+MemoryFile.swift` `funopen` callbacks: Apple imports the prototypes with IUO callback parameters (`(UnsafeMutableRawPointer!, UnsafeMutablePointer<Int8>!, Int32) -> Int32`), Bionic with strictly non-optional pointers. Swift refuses to coerce one `@convention(c)` signature into the other (function pointer types are invariant), so split the existing `funopen` branch into separate Apple and Android sub-branches with the signatures each platform's libc actually expects. The `fopencookie` branch stays unchanged.

## Validation

Built locally on macOS (Xcode 26, Swift 6.3) — clean.

Validated on Android via the [SwiftPorts CI matrix](https://github.com/Cocoanetics/SwiftPorts/pull/3): with this fork pinned in place of upstream `weichsel/ZIPFoundation`, `swift build` on the swift.org Android SDK 6.3 (via [skiptools/swift-android-action@v2](https://github.com/skiptools/swift-android-action)) compiles ZIPFoundation cleanly. The remaining Android failures in that matrix are downstream (libgit2 macros not importing) — none are in ZIPFoundation.

Linux + Windows continue to build green as a regression check.

## Out of scope

- Test target imports — `Tests/ZIPFoundationTests/*` also touch libc directly. Happy to send a follow-up if you want test-target Android coverage too; left out here so the diff stays focused on the library.
- Symlink permissions/atimes on Linux — shipping today as a soft no-op; this PR doesn't try to make them work on non-Apple, just stops them from breaking the Android build.


Sources/ZIPFoundation/Archive+MemoryFile.swift
```diff
@@ -29,12 +34,39 @@ extension Archive {
...
+            // Pre-populate the temp file with whatever bytes the caller
+            // already supplied (read mode); writes mutate the temp file.
+            // Bionic imports `fwrite`'s first arg as non-Optional, so
+            // unwrap before calling.
+            self.data.withUnsafeBytes { buffer in
+                if let base = buffer.baseAddress, buffer.count > 0 {
+                    _ = fwrite(base, 1, buffer.count, result)
+                }
+            }
+            rewind(result)
```
weichsel: While these changes allow the library to compile on platforms without `funopen` support, they introduce two issues:
- Data written to result is never synced back to `self.memoryFile`, so it is never actually read.
- Using a temporary file breaks the in-memory model, undermining performance and violating API expectations.

Given that in-memory archives are a niche feature, a more robust approach may be to disable them on platforms that lack equivalents to `funopen`.

Package.swift
```diff
@@ -8,8 +11,22 @@ let targets: [Target] = [
...
+        dependencies: ["CZLib"],
+        cSettings: [.define("_GNU_SOURCE", to: "1")],
+        // Per-platform link library. Windows vcpkg installs `zlib.lib`
+        // so we link `zlib`; Linux / Android ship `libz.so` aka `-lz`.
+        // Apple takes the canImport(Compression) branch above and
+        // doesn't reach here.
+        linkerSettings: [
+            .linkedLibrary("z", .when(platforms: [.linux, .android])),
+            .linkedLibrary("zlib", .when(platforms: [.windows])),
+        ]),
```
weichsel: We provide multiple SPM manifests. On modern toolchains, SwiftPM will select Package@swift-5.9.swift (for example), so the changes here have no effect in those cases.

weichsel (changes requested):
Thanks for the PR - this is a solid starting point for broader platform support. There are a few build and lint issues, along with some behavioral bugs, that still need to be addressed. Also: What's the main motivation to support Bionic (vs. "vanilla" Android)?

Sources/ZIPFoundation/Archive+MemoryFile.swift
odrobnik: Good catch — both points are correct. The tmpfile path compiled but didn't behave: writes never synced back to `MemoryFile.data`, and even if they had, swapping the in-memory contract for a disk-backed one is wrong.

Took the suggested approach in [aa4ce33](https://github.com/odrobnik/ZIPFoundation/commit/aa4ce33): gated the entire `MemoryFile` machinery (the class, `Archive(data:)`, `makeBackingConfiguration(for:data:)`, and the writing-side memory branches) on `#if swift(>=5.0) && !os(Windows) && !os(Android)`. File-backed archives keep working everywhere; in-memory is just unavailable on platforms without a userspace `FILE*` API. Also dropped the now-redundant Bionic `fwrite` empty-buffer guard since that code never reaches Bionic any more.

Package.swift
odrobnik: Right — the original `Package.swift` change was a no-op on any toolchain that picks the version-specific manifest. Mirrored the per-platform `linkedLibrary` mapping into `Package@swift-5.9.swift` in [26ba786](https://github.com/odrobnik/ZIPFoundation/commit/26ba786) (and dropped the trailing comma SwiftLint flagged in `Package.swift` while I was there).

odrobnik:
Pushed three follow-up commits addressing the review:

- **[aa4ce33](https://github.com/odrobnik/ZIPFoundation/commit/aa4ce33) — disable in-memory archives on Windows / Android.** Replaces the old `tmpfile()` fallback (which silently broke the in-memory contract) with a hard `#if swift(>=5.0) && !os(Windows) && !os(Android)` gate around `MemoryFile`, `Archive(data:)`, `makeBackingConfiguration(for:data:)`, and the writing-side memory branches. File-backed archives keep working everywhere.
- **[26ba786](https://github.com/odrobnik/ZIPFoundation/commit/26ba786) — mirror linker settings in `Package@swift-5.9.swift`.** The original change was a no-op on modern toolchains. Also drops the trailing comma SwiftLint flagged.
- **[ebe5b5f](https://github.com/odrobnik/ZIPFoundation/commit/ebe5b5f) — clear `swiftlint --strict`.** POSIX-shaped names in `PosixCompat+Windows.swift` get a bracketed `swiftlint:disable` (call sites need `mode_t`/`S_IFLNK`/etc. to read identically across platforms); the `FileManager+ZIP.swift` Windows fallbacks moved into a new `FileManager+ZIPWindows.swift`, and the 256-entry `cp437Lookup` moved to a new `Entry+CodePage437.swift` so both files come back under the 400-line cap. `swift build` is warning-free; `swiftlint --strict` reports 0 violations across `Sources` / `Tests` / all `Package*.swift` manifests. (One pre-existing test failure on `testCreateZIP64ArchiveWithLargeSize` reproduces on `upstream/development` — not introduced here.)

> What's the main motivation to support Bionic (vs. "vanilla" Android)?

Bionic *is* Android's libc — every AOSP device, the NDK, and the swift.org Swift Android SDK go through it; there's no separate "vanilla Android" libc. The two module names in the shim (`Android`, `Bionic`) are the same toolchain's libc under different names: the Swift Android SDK currently ships it as `Android`, older / pre-consolidation toolchains exposed it as `Bionic`. Supporting both keeps the fork buildable on either while the upstream Swift renaming settles. The reason explicit imports are needed at all (rather than relying on Foundation re-export the way Darwin / Glibc / ucrt do) is that Foundation on Android doesn't transitively re-export libc — that's the entire purpose of the `#if canImport(Android) … #elseif canImport(Bionic)` shim that appears at the top of the affected files.

weichsel:
Thanks for the follow up.  

There are still some issues that break the Linux build (and also the newly added platforms that are currently not addressed by our CI setup):

- ZIPFoundationFileAttributeTests.swift still calls setSymlinkModificationDate on all platforms, while the helper is now Apple-only. This breaks Linux CI and should be gated accordingly.
- Archive+Deprecated.swift still exposes deprecated init?(data:...) on Windows/Android, while the underlying throwing Archive(data:...) initializer is excluded there. The deprecated initializer needs the same platform gate.
- Memory-archive tests are still registered/referenced under plain swift(>=5.0) in some test files, while excluded on Windows/Android elsewhere. The gates need to be aligned to fix test builds.

We should add a Android CI job to cover this. e.g. via: https://github.com/skiptools/swift-android-action
Same for Windows. Most recent GH Actions runner for that I could find is: https://github.com/compnerd/gha-setup-swift

> the Swift Android SDK currently ships it as Android, older / pre-consolidation toolchains exposed it as Bionic

If I understand your comments and [this Swift Forums post](https://forums.swift.org/t/improving-swift-support-and-interoperability-experience-for-android/71470)￼ correctly, Android is a superset of Android’s C standard library and POSIX support. It also appears to be the modern way to interact with Android from Swift, so I’d prefer using it exclusively instead of Bionic.

> One pre-existing test failure on testCreateZIP64ArchiveWithLargeSize reproduces on upstream/development — not introduced here

Thanks for pointing this out: Fixed on `development`.

odrobnik:
@weichsel I've fixed up the build, please approve workflow run so that we see if that makes everything green.

About android, we can certainly add an Android CI target if you like. But I propose to do that as a a different PR, because it likely will require fixes outside the scope of this one.

odrobnik:
@weichsel ok, all green now

weichsel:
Thanks for fixing builds on the established platforms. However, the issues I outlined above are not addressed for Android and Windows. In particular, the incomplete gating of the memory archive initializer currently prevents the library from building on those platforms, while the remaining issues still break the test targets. (Also, the redundant Bionic-based directives are still in)

odrobnik:
@weichsel apologies for the partial pass on the previous push — went through your list again and pushed the remaining fixes:

- **`Archive+Deprecated.swift` — deprecated `init?(data:)`** ([bf30155](https://github.com/odrobnik/ZIPFoundation/commit/bf30155)) — the deprecated initializer was still exposed on Windows/Android while its body called the now-gated throwing `init(data:...)`. Added `&& !os(Windows) && !os(Android)` to match.
- **Memory-archive test registrations** ([bf30155](https://github.com/odrobnik/ZIPFoundation/commit/bf30155)) — `swift5OnlyTests` in `ZIPFoundationTests.swift` and the in-memory `Archive(data:)` call in `ZIPFoundationArchiveTests.swift` were under plain `#if swift(>=5.0)`; the referenced functions live inside the fully gated `ZIPFoundationMemoryTests.swift`, so on Windows/Android they didn't exist. Aligned both to the same `#if swift(>=5.0) && !os(Windows) && !os(Android)` gate.
- **Redundant Bionic directives** ([e57f0e7](https://github.com/odrobnik/ZIPFoundation/commit/e57f0e7)) — dropped the `#elseif canImport(Bionic) / import Bionic` arms from all nine source files (the eight per-file shims plus the secondary arm in `PosixCompat+Windows.swift`). `canImport(Android)` covers the modern Swift Android SDK; the `Bionic` arm was dead. Updated the remaining inline comments to refer to the Swift Android SDK rather than Bionic.

Re Android/Windows CI: agreed it should land — but I'd like to keep that as a follow-up PR. Wiring up `skiptools/swift-android-action` and `compnerd/gha-setup-swift` against the existing matrix is non-trivial (cross-compile flags, conditional `-lz` link on Android, vcpkg zlib on Windows), and any fallout there would muddy this PR's diff.

Local verification on macOS (Swift 6.3): `swift test -c release -Xswiftc -enable-testing` → 131 tests, 0 failures; `swiftlint --strict` → 0 violations across `Sources` / `Tests` / all `Package*.swift` manifests.

odrobnik:
@weichsel Please permit another workflow run

weichsel:
@odrobnik Thanks for the fixes. I cleaned up a bit and opened a PR (https://github.com/odrobnik/ZIPFoundation/pull/1). Please check if your use case still works after my changes. If so, we can merge.

odrobnik:
I have since replaced ZipFoundation with libarchive, so I cannot test it any more in my code. But I am confident that it would work.

weichsel:
> I have since replaced ZipFoundation with libarchive, so I cannot test it any more in my code. But I am confident that it would work.

OK. Can you please merge: https://github.com/odrobnik/ZIPFoundation/pull/1

odrobnik:
I merged it, but why in my fork?

weichsel:
> I merged it, ...

Thanks. 

> but why in my fork?

Didn't realize you had "Maintainer can modify" on.

